PR #18, opened 2016-12-22
CAF-1998: Using forward path separator regardless of platform.

pcrookshpe:


dermot-hardy:
Since backslash is a valid path-name character in Linux we should also check that the partial references don't contain it, because this fix will end up in them being erroneously changed to forward slashes.

PR #20, opened 2017-01-09
CAF-2007: Error only output workers

glucy:
https://jira.autonomy.com/browse/CAF-2007

Functionality moved provided via minimal core changes after discussions with @dermot-hardy.  The ability to create a worker who does not place success message on its output queue is facilitated by the worker-framework, but large the emphasis is placed on the creator of the worker.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,13 +89,13 @@
         <module>worker-core</module>
         <module>worker-queue-rabbit</module>
         <module>worker-store-cs</module>
-        <module>worker-store-s3</module>
         <module>worker-store-fs</module>
         <module>worker-testing-util</module>
         <module>worker-testing-integration</module>
         <module>worker-framework</module>
         <module>worker-example</module>
         <module>worker-archetype</module>
+        <module>worker-store-s3</module>
     </modules>
 
     <scm>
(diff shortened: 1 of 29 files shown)


conalsmith49 (commented):
For createSuccessResultNoOutput - creating a WorkerResponse with null queuename and null context. - looks good

I did some stepping in to understand how the changes propagate up. In WorkerCore complete(...) method, it checks if(null == targetQueue) and does not publish messages to a queue. - the behaviour we want

acknowledgeTask(String) takes a queueMsgId - a comment in WorkerExecutor ("the reference to the message this task arrived on"). and this appears to be the tag of the input message... this looks ok too.

mulhollandc (commented):
Do we need the "TODO (Greg)" comment prefixes? PLEASE IGNORE THIS (CM).

glucy:
Definitely not @mulhollandc I will remove them immediately.

mulhollandc:
Changes look good to me.

dermot-hardy:
Why are there two `createSuccessResultNoOutput()` methods?  How does it make any sense to have an option for an additional `context` argument when there is no message being published back on to the queue?  Surely the context isn't going anywhere so why provide an option to supply it?

In actual fact, I don't think the `result` argument makes much sense either, given that it is also going nowhere.

Shouldn't we just have a single method which doesn't take any arguments for this?  After all both `result` and `context` can be null and effectively neither are used.

PR #25, opened 2017-01-25
CAF-2033: Allow null to pipe when tracking

conalsmith49:
Review as part of CAF-2033, worker-framework acknowledges tasks with a null queue name i.e. batch worker no output responses.

Job tracking worker changes are here: https://github.hpe.com/caf/job-service/pull/11

The job-service-acceptance-tests now pass. However, there is still an exception that is thrown when the job is being cancelled. This can be seen in a comment in the JIRA here: https://jira.autonomy.com/browse/CAF-2033. This doesn't cause the tests to fail and is not necessarily related to this code change.


glucy (approved):

conalsmith49:
The job service acceptance tests now pass with no exceptions.

glucy (approved):

PR #26, opened 2017-01-26
Created Example container fs

aaronoloughlin:


worker-example/pom.xml
```diff
@@ -46,12 +46,29 @@
...
+        <dependencies>
+            <dependency>
+                <groupId>com.github.workerframework</groupId>
+                <artifactId>worker-framework</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
```
glucy, line 66: Do we really need this Dependency Management section?

worker-example/pom.xml
```diff
@@ -46,12 +46,29 @@
...
 
     <modules>
         <module>worker-example</module>
         <module>worker-example-shared</module>
         <module>worker-example-container</module>
+        <module>worker-example-container-fs</module>
         <module>worker-example-message-builder</module>
     </modules>
```
glucy, line 74: I see below all the folders have been renamed from "worker-example-container" to "worker-example-container-fs" yet we still have a worker-example-container module, is this also correct?

Is this part of having the files for the containers only in one place?

PR #31, opened 2017-02-14
CAF-2088 FS data processing test harness throws too many files open

CJComac:


worker-testing-integration/src/test/java/com/hpe/caf/worker/testing/configuration/FileInputWorkerTaskFactoryTest.java
```diff
@@ -0,0 +1,107 @@
...
+                ClassLoader classLoader = getClass().getClassLoader();
+                String mockFilePath = new File(classLoader.getResource("mockFile.txt").getFile()).getAbsolutePath();
+                fileTestInputData.setInputFile(mockFilePath);
+                fileTestInputData.setUseDataStore(true);
+
+                TestItem testItem = new TestItem("mockFile.txt", fileTestInputData, null);
+                testItem.setInputIdentifier(UUID.randomUUID().toString());
+                try {
+                    testFileInputWorkerTaskFactory.createTask(testItem);
+                } catch (Exception exception) {
```
kploch: Wouldn't it be better to just not catch any exceptions here? I assume the test doesn't throw exceptions if it succeeds?
Or maybe just catch the FileSystemException to output some meaningful error message and then re-throw?

worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/FileInputWorkerTaskFactory.java
```diff
@@ -69,8 +75,9 @@ else if(!Strings.isNullOrEmpty(testItem.getInputData().getStorageReference())){
             if (testItem.getInputData().isUseDataStore()) {
 
                 InputStream inputStream = Files.newInputStream(inputFile);
```
kploch: Wrap it into try-with-resources...

kploch (changes requested):

kploch (approved):

PR #35, opened 2017-02-27
CAF-2306: Add custom validator support

pcrookshpe:
Work-in-progress: In particular, I'm uncomfortable with ValidatorFactory not returning a fresh CustomPropertyValidator instance on each call to create().

### DIFF
--- a/worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/configuration/ValidationSettings.java
+++ b/worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/configuration/ValidationSettings.java
@@ -57,8 +57,8 @@ public class ValidationSettings {
             return this;
         }
 
-        public ValidationSettingsBuilder customValidationProperties(CustomPropertyValidator... customPropertyValidators) {
-            settings.customValidationProperties = Arrays.asList(customPropertyValidators);
+        public ValidationSettingsBuilder customValidators(CustomPropertyValidator... customPropertyValidators) {
+            settings.customValidators = Arrays.asList(customPropertyValidators);
             return this;
         }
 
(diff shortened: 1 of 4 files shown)


worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/configuration/ValidationSettings.java
```diff
@@ -57,6 +57,11 @@ public ValidationSettingsBuilder base64Properties(String... base64Properties) {
             return this;
         }
 
+        public ValidationSettingsBuilder customValidationProperties(CustomPropertyValidator... customPropertyValidators) {
```
kploch: Can you rename the method to something like customValidators? "customValidationProperties" doesn't sound right here....
Other methods have "Properties" in names but these are listing property names.

dermot-hardy (approved):
This looks OK to me although as previously discussed it feels like there is something up with the fact that a `propertyName` has to be supplied to the custom validator.

worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/configuration/ValidationSettings.java
```diff
@@ -57,6 +57,11 @@ public ValidationSettingsBuilder base64Properties(String... base64Properties) {
             return this;
         }
 
+        public ValidationSettingsBuilder customValidators(CustomPropertyValidator... customPropertyValidators) {
+            settings.customValidators = Arrays.asList(customPropertyValidators);
```
kploch, line 61: I believe Arrays.asList will throw an exception if custom argument is null. Add validation for this parameter. I know that other methods don't have it but it's just a code-debt which we don't have to repeat.

kploch (approved):

kploch (approved):

PR #37, opened 2017-03-13
CAF-2375: Alphabetically order yaml testcases

conalsmith49:
Enabling features on the ObjectMapper to sort properties alphabetically and order map entries.


dermot-hardy (approved):

conalsmith49:
I found that `sourceData` was not sorted when testing this change's regeneration of `worker-markup` testcases. This is because the `sourceData` is a `LinkedListMultiMap` which intentionally preserves order and as a result the object mapper will not sort this. 

The reason my yaml sorter application _DOES_ sort it is because it reads everything from an existing testcase, and these are read in as a series of `LinkedHashMap` entries rather than objects. The hash map is then sorted.

The worker framework change appears to sort normal maps, but may not sort sets.

PR #41, opened 2017-03-28
CAF-2334: Update documentation for decoding config

mmaBel:


mmaBel:
@dermot-hardy @andyreidz, changes so far to worker-framework. Have not determined how those API docs were generated with that single index and correct links. Changes to Getting Started documentation were mentioned but I'm not certain what they should be.

PR #42, opened 2017-04-03
CAF-2103: Sub headers

conalsmith49:
Sub heading changes for worker-framework.

Are these suitable or are we looking for more detail in these sub headings?


dermot-hardy:
Like @glucy said, I think these are meant to be more of a slogan.

conalsmith49:
What about something like:

`Cross Platform, Cloud Ready, Microservice Framework`

dermot-hardy (approved):

PR #46, opened 2017-05-22
CAF-2946: Add hash-on-store methods to DataStore

pcrookshpe:
https://jira.autonomy.com/browse/CAF-2946

The surprisingly awkward code in hashStore(InputStream,...) is an attempt to hash the stream while it is being read for storing.

### DIFF
--- a/util-store/src/main/java/com/hpe/caf/util/store/StoreUtil.java
+++ b/util-store/src/main/java/com/hpe/caf/util/store/StoreUtil.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2015-2017 Hewlett Packard Enterprise Development LP.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.util.store;
+
+import com.google.common.hash.Hashing;
+import com.google.common.hash.HashingInputStream;
+import com.hpe.caf.api.worker.DataStore;
+import com.hpe.caf.api.worker.DataStoreException;
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.codec.digest.DigestUtils;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Utility methods for use when storing data assets in DataStores.
+ */
+public final class StoreUtil
+{
+    private StoreUtil() {}
+
+    /**
+     * Store data from a stream, which should be closed by the caller.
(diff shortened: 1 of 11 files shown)


pcrookshpe:
The JIRA ticket https://jira.autonomy.com/browse/CAF-2946 says to extend DataStore but that meant adding default methods to avoid breaking existing implementations of the interface, resulting in the DataStore interface getting more complex than we might have realised when story-pointing the ticket. Let me know if this is acceptable...

worker-api/src/main/java/com/hpe/caf/api/worker/HashStoreResult.java
```diff
@@ -0,0 +1,63 @@
...
+        this.reference = Objects.requireNonNull(reference);
+        this.hash = Objects.requireNonNull(hash);
+    }
+
+    public String getReference()
+    {
+        return reference;
+    }
+
+    public void setReference(final String reference)
```
kploch: I don't think you need setters on this type. And you can make the reference and hash final keeping this type immutable.

kploch (changes requested):

dermot-hardy (changes requested):
Looking at the size of this change, and especially the fact that it has required two extra dependecies to be added to worker-api, which is essentially meant to be an interface-only module, I think we need to move this logic out into another utility module, util-store perhaps.

One utility method would do - we don't really need the byte[] or Path versions.

dermot-hardy (changes requested):
As discussed, try using a `HashingInputStream` instead.  I would expect it to be much simpler.

util-store/src/main/java/com/hpe/caf/util/store/StoreUtil.java
```diff
@@ -0,0 +1,95 @@
...
+     * Store data from a local file. The data will be stored in the supplied DataStore relative to the partial
+     * reference supplied, and the absolute reference of the final location will be returned
+     * along with a SHA-1 hash of the data.
+     * @param dataStore the DataStore into which the data will be stored
+     * @param dataPath path to a file on the local filesystem to store on the remote DataStore
+     * @param partialReference the partial reference, which the data will be stored relative to
+     * @return absolute reference to the stored data, which can be used to retrieve, plus a SHA-1 hash of the stored data
+     * @throws DataStoreException if the data store cannot service the request
+     */
+    public static HashStoreResult hashStore(final DataStore dataStore, final Path dataPath, final String partialReference)
```
dermot-hardy, line 88: This method is reading the file twice.  If it just used the first hashStore() overload that takes a stream it would only be reading the file once.

worker-store-fs/src/test/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStoreTest.java
```diff
@@ -17,10 +17,10 @@
 
 
 import com.hpe.caf.api.ConfigurationException;
-import com.hpe.caf.api.worker.DataStore;
-import com.hpe.caf.api.worker.DataStoreException;
-import com.hpe.caf.api.worker.FilePathProvider;
-import com.hpe.caf.api.worker.ReferenceNotFoundException;
+import com.hpe.caf.api.worker.*;
```
dermot-hardy: You should try to avoid the star import.  It is considered bad practice:
http://checkstyle.sourceforge.net/config_imports.html#AvoidStarImport

dermot-hardy (changes requested):

kploch (approved):

PR #50, opened 2017-06-06
CAF-2952: Priority queues

kploch:


util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/QueueCreator.java
```diff
@@ -90,12 +90,17 @@ public QueueCreator withDeadLetterRoutingKey(String routingKey)
         return this;
     }
 
+    public QueueCreator withMaxPriority(int maxPriority)
+    {
+        this.maxPriority = maxPriority;
```
dermot-hardy: Shouldn't this just do a `propsMap.put()` the same way that the other methods in here do?  And if it did wouldn't it mean that you didn't need all those changes to the `RabbitUtil` class?

util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitUtil.java
```diff
@@ -150,11 +166,35 @@ public static void declareQueue(Channel channel, String queueName, Durability du
...
         Objects.requireNonNull(excl);
         Objects.requireNonNull(act);
         Objects.requireNonNull(queueProps);
-        channel.queueDeclare(queueName, dur == Durability.DURABLE, excl == Exclusivity.EXCLUSIVE, act == EmptyAction.AUTO_REMOVE, queueProps);
+
+        LOG.debug("Declaring worker queue {} with maxPriority={}", queueName, maxPriority);
+
+        Map<String, Object> args = new HashMap<>(queueProps);
+        if (maxPriority > 0) {
+            args.put("x-max-priority", maxPriority);
```
dermot-hardy: Duplicate of previous comment:
Shouldn't this just be done the way the `x-message-ttl`, `x-dead-letter-routing-key`, etc. are done, rather than requiring changes to this class?

worker-api/src/main/java/com/hpe/caf/api/worker/TaskMessage.java
```diff
@@ -86,6 +86,11 @@
      */
     private TaskSourceInfo sourceInfo;
 
+    /**
+     * Task message priority.
+     */
+    private Integer priority;
```
dermot-hardy, line 92: It doesn't seem right to me for the priority to actually be a part of the message.

util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitUtil.java
```diff
@@ -117,7 +120,21 @@ public static Config createLyraConfig(int backoffInterval, int maxBackoffInterva
...
+     * attempt to use a worker queue for the first time.
+     * @param channel the channel to use to declare the queue
+     * @param queueName the name of the worker queue
+     * @param maxPriority the maximum supported priority, pass 0 to disable priority
+     * @throws IOException if the queue is not valid and cannot be used, this is likely NOT retryable
+     */
+    public static void declareWorkerQueue(Channel channel, String queueName, int maxPriority)
+        throws IOException
+    {
+        declareQueue(channel, queueName, Durability.DURABLE, Exclusivity.NON_EXCLUSIVE, EmptyAction.LEAVE_EMPTY, Collections.emptyMap(), maxPriority);
```
dermot-hardy: I think this is possibly the only convenience method needed in here, and we could just create the correct map in here in the first place rather than creating an empty map only for it to be later copied and discarded.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerQueue.java
```diff
@@ -29,6 +29,17 @@
      * @param taskMessage the message to publish
      * @param targetQueue the queue to put the message upon
      * @param headers the map of key/value paired headers to be stamped on the message
+     * @param priority the message priority, greater value means higher priority
+     * @throws QueueException if the message cannot be submitted
+     */
+    void publish(String acknowledgeId, byte[] taskMessage, String targetQueue, Map<String, Object> headers, int priority) throws QueueException;
```
dermot-hardy, line 35: Adding this method is a breaking change anyway, so why bother keeping the original `publish()` method around?  Surely it will no longer be used.  I think it would be better to just have one `publish` method and to add the `priority` to it.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
```diff
@@ -28,15 +28,7 @@
...
-import com.hpe.caf.api.worker.DataStoreException;
-import com.hpe.caf.api.worker.DataStoreProvider;
-import com.hpe.caf.api.worker.ManagedDataStore;
-import com.hpe.caf.api.worker.ManagedWorkerQueue;
-import com.hpe.caf.api.worker.QueueException;
-import com.hpe.caf.api.worker.WorkerException;
-import com.hpe.caf.api.worker.WorkerFactory;
-import com.hpe.caf.api.worker.WorkerFactoryProvider;
-import com.hpe.caf.api.worker.WorkerQueueProvider;
+import com.hpe.caf.api.worker.*;
```
dermot-hardy: Avoid star imports - just list out the specific types that are used, even if there are a lot of them.

worker-api/src/main/java/com/hpe/caf/api/worker/MessagePriorityManagerProvider.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.api.worker;
+
+import com.hpe.caf.api.ConfigurationException;
+import com.hpe.caf.api.ConfigurationSource;
+
+/**
+ * Created by ploch on 30/05/2017.
```
dermot-hardy: No comment would be preferable to this one.  Anybody who cares who created it and when can look it up in source control.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
```diff
@@ -114,7 +106,10 @@ public void run(final WorkerConfiguration workerConfiguration, final Environment
         WorkerThreadPool wtp = WorkerThreadPool.create(workerFactory);
         final int nThreads = workerFactory.getWorkerThreads();
         ManagedWorkerQueue workerQueue = queueProvider.getWorkerQueue(config, nThreads);
-        WorkerCore core = new WorkerCore(codec, wtp, workerQueue, workerFactory, path);
+        MessagePriorityManagerProvider priorityManagerProvider = ModuleLoader.getService(MessagePriorityManagerProvider.class);
```
dermot-hardy, line 119: The way this is written is as if a `MessagePriorityManager` is a separate component, independent of the queueing system, but the actual implementation is in `worker-queue-rabbit` suggesting that it is specific to that queue module. 
 The two things don't go together.

Also I'm not seeing a service registration for `MessagePriorityManagerProvider` - is this file checked in?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -109,7 +113,7 @@ public void setResponse(final WorkerResponse response) {
             response.getTaskStatus(), responseContext,
             response.getQueueReference(), taskMessage.getTracking(),
             new TaskSourceInfo(getWorkerName(responseMessageType), getWorkerVersion()));
-
+        responseMessage.setPriority(priorityManager.getResponsePriority(taskMessage));
```
dermot-hardy, line 116: All of the other properties of the responseMessage were set in the constructor.  Surely this should also be another constructor parameter rather than a separate call afterward?  (Although I still don't understand why the priority needs to be in the task message at all)

worker-queue-rabbit/src/main/java/com/hpe/caf/worker/queue/rabbit/RabbitWorkerQueueConfiguration.java
```diff
@@ -71,6 +71,12 @@
     @Min(1)
     private int retryLimit;
 
+    /**
+     * The maximum message priority supported by this worker queue. 0 to disable priority feature.
+     */
+    @Min(0)
+    @Max(255)
+    private int maxPriority;
```
dermot-hardy, line 79: Is making this an `int` a breaking change?  I'm not 100% how this would serialise is `maxPriority` was not present but we should aim to get it to have the correct default (as we cannot expect that all workers will move to environment variable based configuration immediately).  I would suggest that this should be `Integer` so that it can be `null`, and that the default JavaScript configuration file defaults to `undefined` rather than `255`, and that the code actually handles the default in this case.

worker-default-configs/config/cfg~caf~worker~RabbitWorkerQueueConfiguration.js
```diff
@@ -19,5 +19,6 @@
             || (getenv("CAF_WORKER_BASE_QUEUE_NAME") || getenv("CAF_WORKER_NAME") || "worker") + "-in",
     retryQueue: getenv("CAF_WORKER_RETRY_QUEUE") || undefined,
     rejectedQueue: "worker-rejected",
-    retryLimit: getenv("CAF_WORKER_RETRY_LIMIT") || 10
+    retryLimit: getenv("CAF_WORKER_RETRY_LIMIT") || 10,
+    maxPriority: getenv("CAF_RABBITMQ_MAX_PRIORITY") || 255
```
dermot-hardy: For backward compatibility it would be better if the default was `undefined`, so that the `worker-queue-rabbit` code could take care of defaulting to `255` if the parameter is not specified.

worker-queue-rabbit/src/main/java/com/hpe/caf/worker/queue/rabbit/WorkerPublishQueueEvent.java
```diff
@@ -32,7 +32,19 @@
...
     private final long tag;
     private final Map<String, Object> headerMap;
+    private final int priority;
 
+    /**
+     * Create a new WorkerPublishQueueEvent
+ * @param messageData the raw message data to publish
+     * @param routingKey the routing key to publish the data on
+ * @param ackId the id of a message previously consumed to acknowledge
+ * @param headers the map of key/value paired headers to be stamped on the message
```
dermot-hardy: JavaDoc whitespace screw-up

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -396,7 +394,7 @@ public void complete(final String queueMsgId, final String queue, final TaskMess
                     // **** Normal Worker ****                    
                     // A worker with an input and output queue.
                     byte[] output = codec.serialise(responseMessage);
-                    workerQueue.publish(queueMsgId, output, targetQueue, Collections.emptyMap());
+                    workerQueue.publish(queueMsgId, output, targetQueue, Collections.emptyMap(), responseMessage.getPriority() == null ? 0 : responseMessage.getPriority());
```
dermot-hardy, line 397: You're adding the priority here in the `complete()` method but not adding it in the `forward()` method.  Does that mean the priorities are getting lost when then Job Service is in the mix and the Job Tracking Worker is forwarding messages to the target worker?

worker-queue-rabbit/src/main/java/com/hpe/caf/worker/queue/rabbit/ZeroMessagePriorityManager.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Implementation of MessagePriorityManager which always return zero.
+ * It should be used when priorities are disabled or queue implementation doesn't support them.
+ */
+public class ZeroMessagePriorityManager implements MessagePriorityManager
+{
+    @Override
+    public Integer getResponsePriority(final TaskMessage originalTaskMessage)
+    {
+        return 0;
```
dermot-hardy, line 30: Why zero rather than `null`, since you're using `Integer` rather than `int`?

worker-queue-rabbit/src/main/java/com/hpe/caf/worker/queue/rabbit/MessagePriorityManagerProviderImpl.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package com.hpe.caf.worker.queue.rabbit;
+
+import com.hpe.caf.api.ConfigurationException;
+import com.hpe.caf.api.ConfigurationSource;
+import com.hpe.caf.api.worker.MessagePriorityManager;
+import com.hpe.caf.api.worker.MessagePriorityManagerProvider;
+
+/**
+ * Created by ploch on 30/05/2017.
```
dermot-hardy: Comment

dermot-hardy (changes requested):

dermot-hardy (changes requested):
I made individual comments here and there but isn't there a much easier way this can be done, limiting it to an implementation detail of `util-rabbitmq` and `worker-queue-rabbit` rather than building it out across the framework like this?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
kploch: Yes it is checked in.

util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitUtil.java
kploch: Grand, I missed that. Corrected.

dermot-hardy (approved):

PR #51, opened 2017-06-09
CAF-2852: Adjusted StreamingWorkerThreadPool to handle exceptions…

CJComac:
…returned from workers correctly so that the throwable handler is ran. Adjusted tests to match changes.

### DIFF
--- a/worker-core/src/main/java/com/hpe/caf/worker/core/StreamingWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/hpe/caf/worker/core/StreamingWorkerThreadPool.java
@@ -25,6 +25,9 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.RunnableFuture;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


worker-core/src/main/java/com/hpe/caf/worker/core/StreamingWorkerThreadPool.java
```diff
@@ -119,6 +122,17 @@ public int abortTasks() {
         public void afterExecute(Runnable r, Throwable t) {
             try {
                 super.afterExecute(r, t);
+                if (t == null && r instanceof Future<?>) {
+                    try {
+                        ((Future<?>) r).get();
+                    } catch (CancellationException ce) {
+                        t = ce;
```
dermot-hardy: Seems like `CancellationException` should not trigger the worker to crash.  It looks like it would be triggered if there was a temporary disturbance in the connection to RabbitMQ.

worker-core/src/test/java/com/hpe/caf/worker/core/WorkerCoreTest.java
```diff
@@ -190,7 +190,7 @@ public void testAbortTasks()
     {
         BlockingQueue<byte[]> q = new LinkedBlockingQueue<>();
         Codec codec = new JsonCodec();
-        WorkerThreadPool wtp = WorkerThreadPool.create(2);
+        WorkerThreadPool wtp = WorkerThreadPool.create(2, () -> System.out.println("Hit runnable due to interruption"));
```
dermot-hardy: This change looks dodgy - it suggests that there might be an issue.

dermot-hardy (changes requested):
Please explain what causes an abort in the worker, so that we know whether this code is correct or not

worker-core/src/main/java/com/hpe/caf/worker/core/StreamingWorkerThreadPool.java
CJComac: In the latest commit, I have adjusted this to print a log error when a CancellationException occurs

worker-core/src/test/java/com/hpe/caf/worker/core/WorkerCoreTest.java
CJComac: Removed override of default handler as throwable handler would not get invoked in this test

dermot-hardy (changes requested):
The code looks OK but the tests suggest something is wrong, as you've had to introduce non-default handlers. 
 Why are these handlers being called?

dermot-hardy (approved):

glucy:
@CJComac Dermot has approved this, if you are finished you can "Squash and Merge".

PR #56, opened 2017-06-29
Perf json lzf codec

kploch:


worker-testing-util/pom.xml
```diff
@@ -43,22 +43,22 @@
         <dependency>
             <groupId>com.github.workerframework</groupId>
             <artifactId>worker-configs</artifactId>
-            <version>2.0.0-SNAPSHOT</version>
```
dermot-hardy, line 46: There's no need for the changes to this file, since the maven versions plugin updates the version numbers to keep them in-sync anyway.

standard-worker-container/pom.xml
```diff
@@ -34,7 +34,7 @@
     <dependencies>
         <dependency>
             <groupId>com.github.cafapi.codec</groupId>
-            <artifactId>codec-json</artifactId>
```
dermot-hardy, line 37: Obviously we can't actually merge this into `develop` as it would be a breaking change.

I've created a developer build for it:
http://sou-jenkins2.hpeswlab.net/job/WorkerFramework/view/Developer/

xbreizh:
@dermot-hardy Should we keep or remove that PR?

dermot-hardy:
It's too old - it will need to be looked at fresh.

PR #60, opened 2017-11-24
CAF-2670: Changes to enable support for resetting the 'trackTo' field to null when a job has no sub tasks.

mulhollandc:
https://jira.autonomy.com/browse/CAF-2670

I have introduced changes to the WF to support the completion of job with no subtasks.
This change allows the tracking 'trackTo' field to be set to null. The batch worker will utilise these changes.

Associated PRs for CAF-2670:
Worker-Batch: https://github.com/JobService/worker-batch/pull/6
Worker-GlobFilter: https://github.com/JobService/worker-globfilter/pull/4
Job-Service: https://github.com/JobService/job-service/pull/32

### DIFF
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
@@ -28,6 +28,7 @@ public class WorkerResponse
     private final String messageType;
     private final int apiVersion;
     private final byte[] context;
+    private final boolean resetTrackTo;
 
     /**
      * Create a new WorkerResponse.
@@ -41,6 +42,22 @@ public class WorkerResponse
      */
     public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context)
     {
+        this(queue, status, data, msgType, version, context, false);
+    }
+
+    /**
+     * Create a new WorkerResponse.
+     *
+     * @param queue the reference to the queue that the response data should be put upon. This can be null if no queue is provided
+     * @param status the status of the message the Worker is returning
+     * @param data the serialised task-specific data returned from the Worker internals
+     * @param msgType the task-specific message classifier
+     * @param version the task-specific message API version
+     * @param context the new context to add to the task message, can be null
+     * @param resetTrackTo indicates if the tracking 'trackTo' field should be reset
+     */
+    public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context, boolean resetTrackTo)
+    {
         this.queueReference = queue; // queueReference can be 'null' for a dead end worker and
         // 'null' for a worker who does not send success messages to
         // it's output queue
@@ -49,6 +66,7 @@ public class WorkerResponse
         this.messageType = Objects.requireNonNull(msgType);
         this.apiVersion = version;
         this.context = context;
+        this.resetTrackTo = resetTrackTo;
     }
 
     public TaskStatus getTaskStatus()
@@ -79,5 +97,10 @@ public class WorkerResponse
     public byte[] getContext()
     {
         return context;
+    }
+
+    public boolean getResetTrackTo()
+    {
+        return resetTrackTo;
     }
 }
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
@@ -56,7 +56,7 @@ public class WorkerResponse
      * @param context the new context to add to the task message, can be null
      * @param resetTrackTo indicates if the tracking 'trackTo' field should be reset
      */
-    public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context, boolean resetTrackTo)
+    public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context, final boolean resetTrackTo)
     {
         this.queueReference = queue; // queueReference can be 'null' for a dead end worker and
         // 'null' for a worker who does not send success messages to
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
@@ -28,7 +28,7 @@ public class WorkerResponse
     private final String messageType;
     private final int apiVersion;
     private final byte[] context;
-    private final boolean resetTrackTo;
+    private final String trackTo;
 
     /**
      * Create a new WorkerResponse.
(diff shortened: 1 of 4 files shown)


glucy (approved):

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
```diff
@@ -40,6 +41,22 @@
...
+     *
+     * @param queue the reference to the queue that the response data should be put upon. This can be null if no queue is provided
+     * @param status the status of the message the Worker is returning
+     * @param data the serialised task-specific data returned from the Worker internals
+     * @param msgType the task-specific message classifier
+     * @param version the task-specific message API version
+     * @param context the new context to add to the task message, can be null
+     * @param resetTrackTo indicates if the tracking 'trackTo' field should be reset
+     */
+    public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context, boolean resetTrackTo)
```
dermot-hardy: New parameter should be marked `final`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -143,11 +143,26 @@ private TaskMessage createResponseMessage(final boolean includeTaskContext, fina
...
+        //  Check if the 'trackTo' field needs to be reset.
+        final TrackingInfo trackingInfo;
+        if (response.getResetTrackTo()) {
+            //  Reset trackTo field to null;
+            final TrackingInfo taskMessageTracking = taskMessage.getTracking();
+            trackingInfo = new TrackingInfo(taskMessageTracking.getJobTaskId(),
+                    taskMessageTracking.getStatusCheckTime(),
+                    taskMessageTracking.getStatusCheckUrl(),
+                    taskMessageTracking.getTrackingPipe(),
+                    null);
```
dermot-hardy: - Don't copy all the fields - instead use the copy constructor and then call the `setTrackTo()` mutator to change that one field.  For an example see the `updateTaskId()` method.
- You call `taskMessage.getTracking()` and then use it without checking if it is `null`.  I suspect this will cause a `NullPointerException` for messages that haven't been submitted using the Job Service.

worker-caf/src/main/java/com/hpe/caf/worker/AbstractWorker.java
```diff
@@ -130,9 +130,9 @@ protected final WorkerResponse createSuccessResult(final V result, final byte[]
      *
      * @return a WorkerResponse that represents a success
      */
-    protected final WorkerResponse createSuccessNoOutputToQueue()
+    protected final WorkerResponse createSuccessNoOutputToQueue(boolean resetTrackTo)
```
dermot-hardy: This looks like a breaking change - any worker previously using `createSuccessNoOutputToQueue()` would presumably not work following this change.  Better to add a new method for this.

dermot-hardy (changes requested):
I'm still musing over some of these changes.  I want another review round after this one.

mulhollandc:
Hi @dermot-hardy Thanks for the feedback. I have made the changes.

dermot-hardy (changes requested):
Does this code definitely work?  Looking in `checkForTrackingTermination()` in `WorkerCore.java` it looks like it would cause a `NullPointerException`.  (It is hard to follow the Worker Framework code though so I might have missed something.)

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -143,11 +143,24 @@ private TaskMessage createResponseMessage(final boolean includeTaskContext, fina
 
         final String responseMessageType = response.getMessageType();
 
+        //  Check if the 'trackTo' field needs to be reset.
+        final TrackingInfo trackingInfo;
+        if (response.getResetTrackTo()) {
+            //  Reset trackTo field to null;
+            trackingInfo = taskMessage.getTracking();
+            if (trackingInfo != null) {
+                trackingInfo.setTrackTo(null);
```
dermot-hardy: My previous comment was to use the copy constructor and then call the `setTrackTo()` mutator - you are just using the mutator without calling the copy constructor first.  The copy constructor is here: https://github.com/WorkerFramework/worker-framework/blob/9548a0ed0a2963fd3805430dff5277644c43cb5b/worker-api/src/main/java/com/hpe/caf/api/worker/TrackingInfo.java#L80

dermot-hardy (changes requested):

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerResponse.java
```diff
@@ -40,6 +41,22 @@
...
+     *
+     * @param queue the reference to the queue that the response data should be put upon. This can be null if no queue is provided
+     * @param status the status of the message the Worker is returning
+     * @param data the serialised task-specific data returned from the Worker internals
+     * @param msgType the task-specific message classifier
+     * @param version the task-specific message API version
+     * @param context the new context to add to the task message, can be null
+     * @param resetTrackTo indicates if the tracking 'trackTo' field should be reset
+     */
+    public WorkerResponse(final String queue, final TaskStatus status, final byte[] data, final String msgType, final int version, final byte[] context, final boolean resetTrackTo)
```
dermot-hardy: I think this should be more generic.  Rather than having a `resetTrackTo` parameter that sets the `trackTo` pipe to `null` just allow the `trackTo` pipe to be set.  In principle I can think of cases where you'd want to set this to a value other than `null`.

dermot-hardy (changes requested):

mulhollandc:
Hi @dermot-hardy Additional changes addressed.

worker-caf/src/main/java/com/hpe/caf/worker/AbstractWorker.java
```diff
@@ -132,7 +132,18 @@ protected final WorkerResponse createSuccessResult(final V result, final byte[]
...
+        return createSuccessNoOutputToQueue("");
+    }
+
+    /**
+     * Utility method for creating a WorkerReponse that represents a success, but does not send a message to the worker's output message.
+     *
+     * @param trackTo the tracking 'trackTo' pipe to set
+     * @return a WorkerResponse that represents a success
+     */
+    protected final WorkerResponse createSuccessNoOutputToQueue(final String trackTo)
```
dermot-hardy: This method doesn't make sense, because it hardcodes no target queue but allows a final tracking queue to be specified.  Replace it with a `createTaskCompleteResponse()` method.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -143,17 +143,44 @@ private TaskMessage createResponseMessage(final boolean includeTaskContext, fina
 
         final String responseMessageType = response.getMessageType();
 
+        //  Check if a tracking change is required.
+        final TrackingInfo trackingInfo;
+        if (null != response.getTrackTo()) {
```
dermot-hardy: I believe this code can be simplified by checking for empty string first rather than `null`.

dermot-hardy (changes requested):

mulhollandc:
Hi @dermot-hardy Hopefully final changes here. I have implemented what I hope we agreed on yesterday.

dermot-hardy (approved):

PR #61, opened 2017-12-21
If invalid regex is encountered when validating values, return false …

mmaBel:
…rather than throw an exception.


mmaBel:
Required for document workers (see pull request [here](https://github.com/CAFDataProcessing/worker-document/pull/17)).

worker-testing-integration/src/main/java/com/hpe/caf/worker/testing/validation/ValuePropertyValidator.java
```diff
@@ -16,9 +16,11 @@
 package com.hpe.caf.worker.testing.validation;
 
 import org.glassfish.jersey.internal.util.Base64;
+import org.joda.time.DateTime;
```
dermot-hardy, line 19: We shouldn't be using JodaTime.  The fact that you didn't have to add a dependency to the pom for this is worrying.

dermot-hardy (approved):

PR #63, opened 2018-01-08
CAF-3943: Support asynchronous job updates

mulhollandc:
https://jira.autonomy.com/browse/CAF-3943

Can you review my changes to the worker-framework to support sending messages directly to the Job Tracking Worker (JTW). Summary of changes include the following:

1. New module (i.e. worker-report-updates-jtw) introduced to allow report update tasks to be created and sent to the JTW. The JTW will utilise this module to consume and process the new task message type.
2. Changes to WorkerCallback to facilitate sending report update messages to the JTW.
3. Changes to allow JTW to send message directly to the messaging queue for dependency jobs returned when the job task is complete.
4. New ProgressReportBuffer added to WorkerTaskImpl to store report update messages until buffer needs flushed. Buffer limit is also configurable via enviro variable.
5. Needed to pass message headers through to WorkerTaskImpl to check for rejected message header details.
6. WorkerTaskImpl class needs to utilise Codec class in order to serialise the report updates task data.
7. Testing changes.


pom.xml
```diff
@@ -106,6 +106,7 @@
         <module>worker-archetype</module>
         <module>worker-store-s3</module>
         <module>worker-store-mem</module>
+        <module>worker-report-updates-jtw</module>
```
dermot-hardy: I don't like the name of this module - it shouldn't refer to the Job Tracking Worker.

worker-report-updates-jtw/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <artifactId>worker-framework-aggregator</artifactId>
+        <version>2.4.0-SNAPSHOT</version>
+    </parent>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.github.workerframework</groupId>
+            <artifactId>worker-api</artifactId>
+            <version>2.4.0-SNAPSHOT</version>
+        </dependency>
```
dermot-hardy: Is the `worker-api` dependency actually required?

worker-report-updates-jtw/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <dependency>
+            <groupId>javax.validation</groupId>
+            <artifactId>validation-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
```
dermot-hardy: Do we really need `jackson-databind`?  Would `jackson-annotations` not be sufficient?

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskConstants.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.worker.report.updates.jtw;
+
+public class ReportUpdatesTaskConstants
```
dermot-hardy: Mark this class (and all the classes in this module) `final`.

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskConstants.java
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Identifies the sort of task this message is.
+     */
+    public static final String REPORT_UPDATES_TASK_NAME = "JobTrackingWorkerReportUpdatesTask";
```
dermot-hardy: Don't mention the Job Tracking Worker.  I suggest `TrackingReportTask`.

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+    private String failureSource;
+
+    /**
+     * The failure message in the event of an invalid task or exception.
+     */
+    private String failureMessage;
+
+    public String getJobTaskId() {
+        return taskId;
+    }
```
dermot-hardy: Encapsulation is meaningless in this class and in all the classes in this module as it simply intended to define a serialization format.  Scrap all the accessors and mutators and just make the fields public.  (See https://github.com/CAFDataProcessing/worker-document/tree/v3.2.0/worker-document-shared modules for example).

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+ * Holds task data for a progress report update.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class ReportUpdatesTaskData
+{
+    /**
+     * The identifier of the tracked job task whose progress is being reported.
+     */
+    @NotNull
+    private String taskId;
```
dermot-hardy: Rename this to `jobTaskId` as we want to align with what is in the `TrackingInfo` (i.e. https://github.com/WorkerFramework/worker-framework/blob/v2.3.1/worker-api/src/main/java/com/hpe/caf/api/worker/TrackingInfo.java#L38) and also so as to minimise any potential confusion with the `taskId` on `TaskMessage`.

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * The source of failure in the event of an invalid task or exception.
+     */
+    private String failureSource;
+
+    /**
+     * The failure message in the event of an invalid task or exception.
+     */
+    private String failureMessage;
```
dermot-hardy: Let's group the four failure fields into a failure object.

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+     * The tracked job task's estimated percentage complete.
+     */
+    @Min(0)
+    @Max(100)
+    private int estimatedPercentageCompleted;
+
+    /**
+     * The number of attempted retries for the job task.
+     */
+    private int retries;
```
dermot-hardy: Is this field relevant to all status report types?  If not, then it should be `Integer` rather than `int` so that it can be `null` for status types where it is not used.

Similar question for `estimatedPercentageCompleted` above.  Is it relevant to all status reports?

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+     * The identifier of the tracked job task whose progress is being reported.
+     */
+    @NotNull
+    private String taskId;
+
+    /**
+     * Specifies how the job task should be reported.
+     */
+    @NotNull
+    private ReportUpdatesStatus reportUpdatesStatus;
```
dermot-hardy: The field name will come through in the Json so I'd be inclined to just call it `status` rather than `reportUpdatesStatus`.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -64,4 +64,13 @@
...
     void discard(String queueMsgId);
+
+    /**
+     * Used to send a report update response prior to the final response (when complete should be used instead).
+     *
+     * @param queueMsgId a queue-specific reference for the incoming message that generated the response
+     * @param queue the queue to hold the report update response
+     * @param responseMessage the report update response to put on the queue
+     */
+    void reportUpdate(final String queueMsgId, final String queue, final TaskMessage responseMessage);
```
dermot-hardy: Do we really need the second parameter here?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
```diff
@@ -107,8 +108,8 @@ public void discardTask(final TaskMessage tm, final String queueMessageId) throw
     /**
      * Creates a WorkerTask for the specified message
      */
-    private WorkerTaskImpl createWorkerTask(final String messageId, final TaskMessage taskMessage, boolean poison)
+    private WorkerTaskImpl createWorkerTask(final String messageId, final TaskMessage taskMessage, final boolean poison, final Map<String, Object> headers)
```
dermot-hardy: 138-character guideline breached

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -42,18 +57,22 @@
     private int responseCount;
     private final Object responseCountLock;
     private final SingleResponseMessageBuffer singleMessageBuffer;
+    private final ProgressReportBuffer progressReportBuffer;
     private final AtomicInteger currentSubtaskId;
     private final Semaphore subtasksPublishedSemaphore;
     private final boolean poison;
+    private Codec codec = null;
+    private Map<String, Object> headers;
```
dermot-hardy: Both of these should be `final`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -65,9 +84,18 @@ public WorkerTaskImpl(
...
         this.responseCountLock = new Object();
         this.singleMessageBuffer = new SingleResponseMessageBuffer();
+        this.progressReportBuffer = new ProgressReportBuffer();
         this.currentSubtaskId = new AtomicInteger();
         this.subtasksPublishedSemaphore = new Semaphore(0);
         this.poison = poison;
+        this.headers = headers;
+
+        try {
+            codec = ModuleLoader.getService(Codec.class);
```
dermot-hardy: You updated the various layers to pass through the headers - why not pass through the codec as well?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
         }
     }
 
+    /**
+     * Holds a message along with any headers stamped on the message. This is used to publish report updates.
+     */
+    private final class TaskMessageAndHeaders
+    {
+        private TaskMessage message;
+        private Map<String, Object> headers;
```
dermot-hardy: As discussed elsewhere this object should be removed, but if it was to be kept then both of these fields should be marked `final`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+
+        //  Build up ReportUpdatesTask data to send to tracking pipe.
+        final ReportUpdatesTask reportUpdatesTask = new ReportUpdatesTask();
+        final List<ReportUpdatesTaskData> reportUpdatesTaskDataList = new ArrayList<>();
+
+        //  Iterate through each task message and corresponding headers to generate a progress report update.
+        for (final TaskMessageAndHeaders msgAndHeaders : taskMessageAndHeaders) {
+            //  Retrieve message and header details.
+            final TaskMessage tm = msgAndHeaders.getTaskMessage();
+            final Map<String, Object> headers = msgAndHeaders.getHeaders();
```
dermot-hardy: If you removed this line the code would still compile and would do exactly the same thing, because you've already got headers stored in an instance variable.  With this realization you should see that you can completely remove the `TaskMessageAndHeaders` object.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -123,6 +151,18 @@ public void addResponse(final WorkerResponse response, final boolean includeTask
         final TaskMessage responseMessage = createResponseMessage(includeTaskContext, response);
 
         singleMessageBuffer.add(responseMessage);
+
+        //  Add report update task message to progress report buffer.
+        progressReportBuffer.add(responseMessage, headers);
```
dermot-hardy: Passing headers into the progress report buffer shouldn't be needed.  See comment in the `getReportUpdatesTask()` function.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
...
+
+    /**
+     * Holds the progress report update messages that needs to be forwarded on to the tracking pipe.
+     */
+    private final class ProgressReportBuffer
+    {
+        private List<TaskMessageAndHeaders> buffer;
+        private final Object bufferLock;
+        private boolean isBuffering;
+        private int bufferLimit;
```
dermot-hardy: The `buffer` and `bufferLimit` fields should both be `final`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
...
+            this.buffer = new ArrayList<>();
+            this.bufferLock = new Object();
+            this.isBuffering = true;
+
+            final String reportUpdatesBufferLimit = System.getenv("CAF_WORKER_REPORT_UPDATES_BUFFER_LIMIT");
+            if (null == reportUpdatesBufferLimit) {
+                // Default buffer limit to 5 if the environment variable is not present.
+                this.bufferLimit = 5;
+            } else {
+                this.bufferLimit = Integer.parseInt(reportUpdatesBufferLimit);
```
dermot-hardy: We need to catch the potential `NumberFormatException` that would be thrown if the environment variable was set to a non-integer.  Perhaps just log a warning and go with the default size.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
...
+        private boolean isBuffering;
+        private int bufferLimit;
+
+        public ProgressReportBuffer()
+        {
+            this.buffer = new ArrayList<>();
+            this.bufferLock = new Object();
+            this.isBuffering = true;
+
+            final String reportUpdatesBufferLimit = System.getenv("CAF_WORKER_REPORT_UPDATES_BUFFER_LIMIT");
```
dermot-hardy: Pull this logic out into a `private static int getBufferLimit() { ... }` function.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
...
+            synchronized (bufferLock) {
+                //  If buffer limit has exceeded, then it needs to be flushed.
+                if ((buffer.size() >= bufferLimit) || (!isBuffering)) {
+                    bufferContentsToPublish = new ArrayList<>(buffer);
+                    buffer.clear();
+                }
+
+                //  If buffering is enabled, then add message and corresponding headers to the buffer if it does not
+                //  already exist.
+                if (isBuffering && (!buffer.contains(msgAndHeaders))) {
```
dermot-hardy: How could the buffer possibly contain this message?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +427,110 @@ private void addOrFlush(final TaskMessage newResponse)
...
+                    buffer.add(msgAndHeaders);
+                }
+            }
+
+            //  Publish the messages currently in the buffer to be published.
+            if (bufferContentsToPublish != null && bufferContentsToPublish.size() > 0) {
```
dermot-hardy: I think it might be clearer to make this check inside the function.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                ReportUpdatesTaskConstants.REPORT_UPDATES_TASK_API_VER, reportUpdatesTaskData, TaskStatus.NEW_TASK,
+                Collections.<String, byte[]>emptyMap(), trackingPipe);
+
+        //  Publish the task message comprising the report updates.
+        workerCallback.reportUpdate(messageId, trackingPipe, responseMessage);
+
+        // TODO - do we need a semaphore here?
+    }
+
+    private String getTrackingPipe(final List<TaskMessageAndHeaders> taskMessageAndHeaders)
```
dermot-hardy: This method looks like it could be `static`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+    {
+        //  Return the first tracking pipe. All task messages are expected to comprise the same
+        //  tracking pipe.
+        return taskMessageAndHeaders.get(0).getTaskMessage().getTracking().getTrackingPipe();
+    }
+
+    private ReportUpdatesTask getReportUpdatesTask(final List<TaskMessageAndHeaders> taskMessageAndHeaders) {
```
dermot-hardy: I think you should rename this to `createReportUpdatesTask()`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+     * Used to publish progress report update messages onto the tracking pipe.
+     */
+    private void publishReportUpdates(final List<TaskMessageAndHeaders> reportUpdates)
+    {
+        //  Make a note of the tracking pipe where progress report updates are to be sent.
+        final String trackingPipe = getTrackingPipe(reportUpdates);
+
+        //  Build up a ReportUpdatesTask comprising a list of progress report updates to send.
+        final ReportUpdatesTask reportUpdatesTask = getReportUpdatesTask(reportUpdates);
+        if (reportUpdatesTask.getTasks() == null || reportUpdatesTask.getTasks().size() == 0) {
```
dermot-hardy: Use `.isEmpty()` rather than `.size() == 0`.  (Although I'm not sure why you really even need to check this here.)

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+        }
+
+        //  Create a task message comprising the progress report updates.
+        final TaskMessage responseMessage = new TaskMessage(
+                UUID.randomUUID().toString(), ReportUpdatesTaskConstants.REPORT_UPDATES_TASK_NAME,
+                ReportUpdatesTaskConstants.REPORT_UPDATES_TASK_API_VER, reportUpdatesTaskData, TaskStatus.NEW_TASK,
+                Collections.<String, byte[]>emptyMap(), trackingPipe);
+
+        //  Publish the task message comprising the report updates.
+        workerCallback.reportUpdate(messageId, trackingPipe, responseMessage);
```
dermot-hardy: Why pass the tracking pipe here instead of getting it from the message?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+
+        // TODO - do we need a semaphore here?
```
dermot-hardy: Let's come back to this one.

worker-report-updates-jtw/src/main/java/com/hpe/caf/worker/report/updates/jtw/ReportUpdatesTaskData.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+import java.util.Date;
+import java.util.Objects;
+
+/**
+ * Holds task data for a progress report update.
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class ReportUpdatesTaskData
```
dermot-hardy: I think we should rename this object because it would be easily confused with the message task data whereas it is actually a report pertaining to a particular task.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                reportUpdatesTaskData.setFailureId(taskStatus.toString());
+                reportUpdatesTaskData.setFailureTime(new Date());
+                reportUpdatesTaskData.setFailureSource(getWorkerName(tm));
+                final byte[] taskData = tm.getTaskData();
+                if (taskData != null) {
+                    reportUpdatesTaskData.setFailureMessage(new String(taskData, StandardCharsets.UTF_8));
+                }
+                //  Task should be reported as rejected.
+                reportUpdatesTaskData.setReportUpdatesStatus(ReportUpdatesStatus.Rejected);
+            } else {
```
dermot-hardy: How do you get into this `else` branch?  You seem to have covered off all possible task statuses in the `if` statements.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                    reportUpdatesTaskData.setEstimatedPercentageCompleted(0);
+                }
+            } else if (taskStatus == TaskStatus.RESULT_EXCEPTION || taskStatus == TaskStatus.INVALID_TASK) {
+                //  Failed to execute job task. Configure failure details to be reported.
+                reportUpdatesTaskData.setFailureId(taskStatus.toString());
+                reportUpdatesTaskData.setFailureTime(new Date());
+                reportUpdatesTaskData.setFailureSource(getWorkerName(tm));
+                final byte[] taskData = tm.getTaskData();
+                if (taskData != null) {
+                    reportUpdatesTaskData.setFailureMessage(new String(taskData, StandardCharsets.UTF_8));
```
dermot-hardy: I think this is right for `RESULT_EXCEPTION` but is it also right for `INVALID_TASK`?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                }
+                //  Task should be reported as rejected.
+                reportUpdatesTaskData.setReportUpdatesStatus(ReportUpdatesStatus.Rejected);
```
dermot-hardy: Is `Rejected` correct for this?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +615,149 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                    reportUpdatesTaskData.setReportUpdatesStatus(ReportUpdatesStatus.Retry);
+                }
+            }
+
+            //  Add ReportUpdatesTaskData to list.
+            reportUpdatesTaskDataList.add(reportUpdatesTaskData);
+        }
+
+        // Add progress report update task data to ReportUpdatesTask and return.
+        if (reportUpdatesTaskDataList.size() > 0) {
```
dermot-hardy: Don't you want to do this always, given that you've marked the `tasks` field as `@NotNull`?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -221,6 +264,9 @@ public void setResponse(final InvalidTaskException invalidTaskException)
             taskMessage.getTracking(),
             new TaskSourceInfo(getWorkerName(taskClassifier), getWorkerVersion()));
 
+        //  Add report update task message to progress report buffer.
+        progressReportBuffer.add(invalidResponse, headers);
```
dermot-hardy: I think it would be better to make the `progressReportBuffer.add()` calls from the `publishSubtask()` and `completeResponse()` methods, rather than from `addResponse()` and the two `setResponse()` methods.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -486,6 +484,30 @@ public void discard(String queueMsgId)
             stats.incrementTasksDiscarded();
         }
 
+        @Override
+        public void reportUpdate(final String queueMsgId, final String queue, final TaskMessage responseMessage)
```
dermot-hardy: `responseMessage` is a confusing variable name as this is not the response.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -486,6 +484,30 @@ public void discard(String queueMsgId)
...
+        {
+            Objects.requireNonNull(queueMsgId);
+            Objects.requireNonNull(queue);
+            Objects.requireNonNull(responseMessage);
+            LOG.debug("Sending report update message to queue {} for task {} (message id: {})", queue, responseMessage.getTaskId(), queueMsgId);
```
dermot-hardy: Is this log message correct?  Is it for the task that it says it is for?

dermot-hardy (changes requested):

pom.xml
mulhollandc: OK but I thought you requested an implementation specific folder for report updates, similarly to 's3' for worker-store.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
mulhollandc: If we don't supply the queue (i.e tracking pipe) how do we know which queue to publish the message on?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
mulhollandc: I agree. I originally had 'reportUpdateMessage' and changed it. Will revert back.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
mulhollandc: Yes could do that.

mulhollandc:
Hi @dermot-hardy Thanks for your feedback. I have hopefully addressed all in my latest changes.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -64,4 +64,12 @@
      * @param queueMsgId the id of the task's queue message to discard
      */
     void discard(String queueMsgId);
+
+    /**
+     * Used to send a report update message prior to the final response (when complete should be used instead).
```
dermot-hardy: Is this description correct?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -486,6 +484,30 @@ public void discard(String queueMsgId)
             stats.incrementTasksDiscarded();
         }
 
+        @Override
+        public void reportUpdate(final String queueMsgId, final TaskMessage reportUpdateMessage)
+        {
+            Objects.requireNonNull(queueMsgId);
+            Objects.requireNonNull(reportUpdateMessage);
+            LOG.debug("Sending report update message to queue {} for task {})", reportUpdateMessage.getTo(),
+                    reportUpdateMessage.getTaskId());
```
dermot-hardy: I still think this log message is not correct.  Won't the task id be a random guid rather than the task the log message says it is for?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
```diff
@@ -64,12 +65,13 @@ public WorkerExecutor(
      *
      * @param tm the task message
      * @param queueMessageId the reference to the message this task arrived on
+     * @param headers the map of key/value paired headers to be stamped on the message
      * @throws TaskRejectedException if the WorkerFactory indicates the task cannot be handled at this time
      */
-    public void executeTask(final TaskMessage tm, final String queueMessageId, boolean poison)
+    public void executeTask(final TaskMessage tm, final String queueMessageId, final boolean poison, final Map<String, Object> headers, final Codec codec)
```
dermot-hardy: 138-character guideline breached

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +418,70 @@ private void addOrFlush(final TaskMessage newResponse)
         }
     }
 
+    /**
+     * Holds the progress report update messages that needs to be forwarded on to the tracking pipe.
+     */
+    private final class ProgressReportBuffer
+    {
+        private List<TaskMessage> buffer;
```
dermot-hardy: Should be `final`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +572,188 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+        final String bufferLimitEnv = System.getenv("CAF_WORKER_REPORT_UPDATES_BUFFER_LIMIT");
+        if (null == bufferLimitEnv) {
+            // Default buffer limit to 5 if the environment variable is not present.
+            bufferLimit = 5;
+        } else {
+            try {
+                bufferLimit = Integer.parseInt(bufferLimitEnv);
+            } catch (final NumberFormatException nfe) {
+                //  Log the environment variables does not contain a parsable int and return default value instead.
+                LOG.debug("CAF_WORKER_REPORT_UPDATES_BUFFER_LIMIT does not contain a parsable int.");
```
dermot-hardy: Make this a warning rather than a debug message.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -462,4 +572,188 @@ private void updateTaskId(final TaskMessage responseMessage, final int subtaskId
...
+                    trackingReport.status = TrackingReportStatus.Retry;
+                }
+            }
+
+            //  Add tracking report to list.
+            trackingReports.add(trackingReport);
+        }
+
+        //  Add list of tracking reports to tracking report task.
+        if (trackingReports.size() > 0) {
```
dermot-hardy: Don't you want to do this always, given that you've marked the `trackingReports` field as `@NotNull`?

In fact I'd be inclined to move the `final TrackingReportTask trackingReportTask = new TrackingReportTask();` line down here also, as it is not needed when you are building up the tracking reports.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +418,70 @@ private void addOrFlush(final TaskMessage newResponse)
...
+        /**
+         * Flush the buffer and turn it off. Any subsequent calls to add() will not be buffered.
+         */
+        public void flush()
+        {
+            isBuffering = false;
+            addOrFlush(null);
+        }
+
+        private void addOrFlush(final TaskMessage taskMessage)
```
dermot-hardy, line 463: I think there is a logic error in this function.  The comment on `flush()` says:

> Flush the buffer and turn it off. Any subsequent calls to add() will not be buffered.

But if there are subsequent calls to `add()` then I think what will actually happen is that the messages will be lost, rather than being published immediately without being buffered.

Looking at the calling code I can see that the semaphore is already ensuring that all `add()` calls come before the `flush()` call anyway, so this is not a real issue the way it is in the `SingleResponseMessageBuffer`, which is not being protected by the semaphore.  What this means is that you could just disallow `add()` calls after `flush()` and make that part of the definition of how this class must be used.

dermot-hardy (changes requested):
A few minor issues - I thought the `ProgressReportBuffer` one was important but then I looked at the calling code and realized that it can't really happen.

mulhollandc:
Hi @dermot-hardy further changes made to address your latest review comments including one to the ProgressReportBuffer to ensure that any subsequent add() calls are not lost when buffering has been disabled.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -381,6 +418,76 @@ private void addOrFlush(final TaskMessage newResponse)
...
+                }
+
+                //  If buffering is enabled, then add message to the buffer if it does not already exist.
+                if (isBuffering) {
+                    buffer.add(taskMessage);
+                } else {
+                    //  Ensure any subsequent calls to add() when buffering is disabled results in the task being
+                    //  immediately published and not lost.
+                    if (taskMessage != null) {
+                        bufferContentsToPublish.add(taskMessage);
```
dermot-hardy, line 481: Technically there's a possibility that this could cause a batch one larger than the normal maximum (if the buffer had just reached the maximum size and then on another thread `flush()` was called), but I'm not worried about it.

dermot-hardy (approved):
Please also update the version number to 3.0.0 as I think we have to consider this to be a breaking change, and update the release notes to describe the feature and the break.

mulhollandc:
Hi @dermot-hardy. Are you suggesting moving JobService or WF up to 3.0.0? If WF and I move to 3.0.0-SNAPSHOT, then I assume all sub modules need also moved to 3.0.0-SNAPSHOT? Should I also update all other projects that have a dependency on WF 2.4.0-SNAPHOT now or park for later? Finally can you help me describe the breaking change as I'm not sure myself? Thanks.

dermot-hardy:
Use the “Change Version” functionality on the build machine to update the version:

[cid:image001.png@01D38BAE.90AF8A30]

Leave the other projects for later.  They can be done before they are released.

“Finally can you help me describe the breaking change as I'm not sure myself?”
Just state that the 2.3.0 Job Service and below is no longer supported.

Also, raise a separate Jira to re-introduce support for the 2.3.0 Job Service.  If we can get that done then we can put this back to 2.4.0-SNAPSHOT.

mulhollandc:
POM and release note changes made.

dermot-hardy (approved):

PR #66, opened 2018-05-18
SCMOD-4072: Updated to use latest version of dropwizard

Anthony-McGreevy:
https://jira.autonomy.com/browse/SCMOD-4072

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015-2017 EntIT Software LLC, a Micro Focus company.
+    Copyright 2018-2017 EntIT Software LLC, a Micro Focus company.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2018-2017 EntIT Software LLC, a Micro Focus company.
+    Copyright 2015-2017 EntIT Software LLC, a Micro Focus company.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/pom.xml
+++ b/pom.xml
@@ -108,6 +108,21 @@
         <module>worker-tracking-report</module>
     </modules>
 
+    <dependencyManagement> 
+        <dependencies>
+            <dependency> 
+                <groupId>io.dropwizard</groupId> 
+                <artifactId>dropwizard-testing</artifactId> 
+                <version>1.3.2</version> 
+            </dependency>
+            <dependency>
+                <groupId>io.dropwizard</groupId>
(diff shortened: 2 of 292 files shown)


Anthony-McGreevy:
**TEST BUILD:** http://sou-jenkins2.hpeswlab.net/job/WorkerFramework/view/Developer/job/WorkerFramework~worker-framework~SCMOD-4072~CI/
**JIRA:** https://jira.autonomy.com/browse/SCMOD-4072

Anthony-McGreevy:
**TEST BUILD:** http://sou-jenkins2.hpeswlab.net/job/WorkerFramework/view/Developer/job/WorkerFramework~worker-framework~SCMOD-4072~CI/

docs/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright 2015-2017 EntIT Software LLC, a Micro Focus company.
+    Copyright 2018-2017 EntIT Software LLC, a Micro Focus company.
```
dermot-hardy: Change the project inception year in `caf-common-parent` back to 2015.  It should be the year that the project started rather than the year that the module itself was created.  (Or you could override the inception year in the aggregator to set it to 2015).

pom.xml
```diff
@@ -108,6 +108,30 @@
...
+                <groupId>io.dropwizard</groupId>
+                <artifactId>dropwizard-bom</artifactId>
+                <version>1.3.2</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>com.github.cafapi</groupId>
+                <artifactId>caf-common</artifactId>
+                <version>1.16.0-SNAPSHOT</version>
```
dermot-hardy: Never reference the SNAPSHOT version unless you're using something from it.  By referencing it you are saying that the latest released version is no good for you, that this new SNAPSHOT version has functionality that you need, and that it will have to be released before you are able to release this.  I know that it not true in this case since nothing has changed in `caf-common` since it's last release.

pom.xml
```diff
@@ -108,6 +108,30 @@
         <module>worker-tracking-report</module>
     </modules>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.hibernate</groupId>
+                <artifactId>hibernate-validator</artifactId>
+                <version>5.2.5.Final</version>
+            </dependency>
```
dermot-hardy, line 117: I'm sure it's correct but as a matter of interest why does this one need to be specifically called out?

dermot-hardy (changes requested):

pom.xml
Anthony-McGreevy: There is a conflict between the versions between the two imported dependency management sections (caf-common(5.1.3.Final) and dropwizard-bom (5.4.2.Final)). If we use the version from either one of these the other no longer works. This version is the middle ground that works for both, I needed to specify it here so that it took precedence over the imported versions.

Anthony-McGreevy:
@dermot-hardy I have updated this PR to reflect your comments, can you look at this again and merge if happy with it.

dermot-hardy (approved):

PR #68, opened 2018-06-04
SCMOD-3294: Workaround Spring dependencies being excluded by default

dermot-hardy:
https://jira.autonomy.com/browse/SCMOD-3294


davidegph (approved):
Everything seems fine.

PR #75, opened 2018-08-24
SCMOD-4898: Add option to not send no progress messages

omkumar-mariappan:
https://jira.autonomy.com/browse/SCMOD-4898


worker-api/src/main/java/com/hpe/caf/api/worker/WorkerConfiguration.java
```diff
@@ -43,4 +45,12 @@ public void setWorkerVersion(String workerVersion)
...
         this.workerVersion = workerVersion;
     }
+
+    public boolean isTrackProgressMessages() {
+        return trackProgressMessages;
+    }
+
+    public void setTrackProgressMessages(boolean trackProgressMessages) {
+        this.trackProgressMessages = trackProgressMessages;
+    }
```
dermot-hardy: I know you talked to me about this but on reflection I don't think we should change this class.  Changing it means that we will have to update all the WorkerConfiguration.js files in each container.

Let's just update the code to use `System.getenv("CAF_WORKER_DISABLE_ZERO_PROGRESS_REPORTING")` from the code instead.  Overall it will reduce the number of changes that we need to make.

worker-example/worker-example-container/src/main/config/cfg~caf~worker~ExampleWorkerConfiguration.js
```diff
@@ -16,6 +16,7 @@
 ({
     workerName: "worker-example",
     workerVersion: "${project.version}",
+    trackProgressMessages: getenv("CAF_WORKER_TRACK_PROGRESS_MESSAGES"),
```
dermot-hardy: The above change will mean that we don't need changes to the JavaScript files for all the workers.

dermot-hardy (changes requested):

omkumar-mariappan:
@dermot-hardy can you check it one more time?

dermot-hardy (approved):

PR #88, opened 2019-07-17
SCMOD-6859: initial commit for health check on channel closures

gaana-vishwnath:
JIRA:
 https://portal.digitalsafe.net/browse/SCMOD-6859

Build: http://sou-jenkins2.hpeswlab.net/job/WorkerFramework/view/Developer/job/WorkerFramework~worker-framework~SCMOD-6859~CI/

### DIFF
--- a/util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitConsumer.java
+++ b/util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitConsumer.java
@@ -34,7 +34,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
  */
 public abstract class RabbitConsumer<T> extends EventPoller<T> implements Consumer
 {
-    private static final Logger LOG = LoggerFactory.getLogger(RabbitConsumer.class);    
+    private static final Logger LOG = LoggerFactory.getLogger(RabbitConsumer.class);
     private final AtomicBoolean channelActive; 
     /**
      * Create a new RabbitConsumer.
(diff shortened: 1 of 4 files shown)


util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitConsumer.java
```diff
@@ -34,7 +35,7 @@
 public abstract class RabbitConsumer<T> extends EventPoller<T> implements Consumer
 {
     private static final Logger LOG = LoggerFactory.getLogger(RabbitConsumer.class);
-
+    private final AtomicBoolean channelActive;
```
andyreidz: This is a consumer not a channel, so consumerActive would be a better choice

andyreidz (commented):
Taking a closer look at this I'm not sure it's the right design choice.

Starting from RabbitWorkerQueue line 100 you can see that a cosumer tag is returned from incommingChannel.basicConsume. The two arguments are the queue name and the consumer object for that queue.  That means there is the potential for a single consumer to be used for multiple queues.

The PR proposes making isActive false when a consumer cancel event is received for any consumer tag. Should the consumer object have been used for multiple queues this would invalidate it for all of them. 

At this point the worker framework does not consume from multiple queues, but there's always the potential for this to be needed in the future.

Possibly I'm worrying needlessly about that happening, I just want to mention it.

            consumer = new DefaultRabbitConsumer(consumerQueue, consumerImpl);
            WorkerPublisherImpl publisherImpl = new WorkerPublisherImpl(outgoingChannel, metrics, consumerQueue, confirmListener);
            publisher = new EventPoller<>(2, publisherQueue, publisherImpl);
            declareWorkerQueue(incomingChannel, config.getInputQueue(), config.getMaxPriority());
            declareWorkerQueue(outgoingChannel, config.getRetryQueue(), config.getMaxPriority());
            synchronized (consumerLock) {
                consumerTag = incomingChannel.basicConsume(config.getInputQueue(), consumer);
            }

gaana-vishwnath:
discarding the implementation.

PR #89, opened 2019-07-24
SCMOD-7272- updated the publish confirm count for retry queue

gaana-vishwnath:
https://portal.digitalsafe.net/browse/SCMOD-7272


dermot-hardy (commented):
I don't get it.  Why would you increment the response count when there hasn't been an additional response?

dermot-hardy (approved):
This seems very asymmetric.  I don't understand why it is not just calling the `RabbitWorkerQueue` methods, but since that is pre-existing this change is OK.

PR #94, opened 2019-10-14
SCMOD-7766: Updated to extend timeout

Anthony-McGreevy:
https://portal.digitalsafe.net/browse/SCMOD-7766


Anthony-McGreevy:
@dermot-hardy Seems like that test had a hardcoded time out of a second. Since the timeout is a max wait time and it will return earlier if the count reaches 0 as it's supposed to I have increased it to 5 seconds, this should be more than enough time to account for the occasional slow down in our build environment.

dermot-hardy (approved):

PR #99, opened 2020-03-13
Fix Bower artifactory Path

rahul-kulkarni73:


dermot-hardy (approved):

dermot-hardy:
I don't know why this was reverted, but I've re-applied it.

PR #108, opened 2020-07-07
SCMOD-9102: Update worker-store-fs healthcheck

rorytorneymf:
Jira: https://portal.digitalsafe.net/browse/SCMOD-9102

Previously, the worker-store-fs healthcheck always returned HEALTHY. Instead, the healthcheck now checks if the `dataDir` directory exists, and will return UNHEALTHY if `dataDir` does not exist, or is not accessible after a configurable period of time: `CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS`.

### DIFF
--- a/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
+++ b/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
@@ -15,8 +15,7 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
-    dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS")
-            || 10,
+    dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS") || undefined,
     outputBufferSize: getenv("CAF_WORKER_DATASTORE_OUTPUT_BUFFER_SIZE")
             || getenv("CAF_WORKER_DATASTORE_BUFFER_SIZE") || undefined
 });
--- a/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
+++ b/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
@@ -15,6 +15,8 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
+    dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS")
+            || 10,
     outputBufferSize: getenv("CAF_WORKER_DATASTORE_OUTPUT_BUFFER_SIZE")
             || getenv("CAF_WORKER_DATASTORE_BUFFER_SIZE") || undefined
 });
--- a/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
+++ b/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
@@ -65,8 +65,6 @@ public class FileSystemDataStore implements ManagedDataStore, FilePathProvider, 
         throws DataStoreException
     {
         dataStorePath = FileSystems.getDefault().getPath(config.getDataDir());
-        healthcheck = new FileSystemDataStoreHealthcheck(dataStorePath);
-        healthcheckTimeout = Duration.ofSeconds(config.getDataDirHealthcheckTimeoutSeconds());
         if (!Files.exists(dataStorePath)) {
             try {
                 Files.createDirectory(dataStorePath);
@@ -74,7 +72,8 @@ public class FileSystemDataStore implements ManagedDataStore, FilePathProvider, 
                 throw new DataStoreException("Cannot create data store directory", e);
             }
         }
-
+        healthcheck = new FileSystemDataStoreHealthcheck(dataStorePath);
+        healthcheckTimeout = Duration.ofSeconds(config.getDataDirHealthcheckTimeoutSeconds());
         outputBufferSize = getOutputBufferSize(config);
 
         LOG.debug("Initialised");
--- a/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
+++ b/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
@@ -16,6 +16,7 @@
 package com.hpe.caf.worker.datastore.fs;
 
 import com.hpe.caf.api.HealthResult;
+import com.hpe.caf.api.HealthStatus;
 import com.hpe.caf.api.worker.*;
 import org.apache.commons.io.output.ProxyOutputStream;
 import org.slf4j.Logger;
@@ -26,8 +27,16 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.*;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Consumer;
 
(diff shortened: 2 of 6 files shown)


worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
```diff
@@ -15,6 +15,8 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
+    dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS")
+            || 10,
```
dermot-hardy: I think most components will use this default configuration file but unfortunately there are some that bundle their own instead.  Here are a few that I found; there may be more:
- [Verity/processing-coordinator/.../cfg\~darwin\~processing-coordinator\~FileSystemDataStoreConfiguration.js](https://github.houston.softwaregrp.net/Verity/processing-coordinator/blob/v2.4.0/processing-coordinator-container/src/main/config/cfg~darwin~processing-coordinator~FileSystemDataStoreConfiguration.js)
- [Verity/view/.../cfg\~darwin\~view\~FileSystemDataStoreConfiguration.js](https://github.houston.softwaregrp.net/Verity/view/blob/v2.4.1/view-node/src/main/config/cfg~darwin~view~FileSystemDataStoreConfiguration.js)

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -60,7 +72,8 @@ public FileSystemDataStore(final FileSystemDataStoreConfiguration config)
                 throw new DataStoreException("Cannot create data store directory", e);
             }
         }
-
+        healthcheck = new FileSystemDataStoreHealthcheck(dataStorePath);
+        healthcheckTimeout = Duration.ofSeconds(config.getDataDirHealthcheckTimeoutSeconds());
```
dermot-hardy: I think that if the configuration isn't set that you will get a `NullPointerException` here.  Obviously this won't usually happen because you've adjusted the default configuration but as per the previous comment I'm not sure that you can always rely on that being used.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -45,6 +54,9 @@
     private final AtomicInteger numDx = new AtomicInteger(0);
     private final DataStoreMetricsReporter metrics = new FileSystemDataStoreMetricsReporter();
     private static final Logger LOG = LoggerFactory.getLogger(FileSystemDataStore.class);
+    private final Callable<HealthResult> healthcheck;
+    private final Duration healthcheckTimeout;
+    private final ExecutorService healthcheckExecutor = Executors.newSingleThreadExecutor();
```
dermot-hardy: It's OK since there are already other variables being initialized in this way but just for the record the initialization should be being done in the constructor.

worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
dermot-hardy: My advice w.r.t. this btw is to fallback to `undefined` the same way `outputBufferSize` is doing, and then to put the default in the code, again like [this](https://github.com/WorkerFramework/worker-framework/blob/v3.3.0/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java#L348).  It will save you having to worry about components which don't use the default configuration file.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStoreHealthcheck.java
```diff
@@ -0,0 +1,43 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.worker.datastore.fs;
+
+import com.hpe.caf.api.HealthResult;
+import com.hpe.caf.api.HealthStatus;
+import java.nio.file.*;
```
dermot-hardy: Don't use star imports.
```suggestion
import java.nio.file.Files;
import java.nio.file.Path;
```

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStoreHealthcheck.java
```diff
@@ -0,0 +1,43 @@
...
+{
+    private final Path dataStorePath;
+
+    public FileSystemDataStoreHealthcheck(final Path dataStorePath)
+    {
+        this.dataStorePath = dataStorePath;
+    }
+
+    @Override
+    public HealthResult call() throws Exception
```
dermot-hardy: Don't declare `throws Exception` if you're not throwing one.
```suggestion
    public HealthResult call()
```

dermot-hardy (changes requested):

worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
rorytorneymf: Using the same approach as `outputBufferSize `uses now:

1. Fallback to undefined in default config file
2. Setting default value in code to handle components that don't use the default conf file

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
rorytorneymf: See previous comment

dermot-hardy (approved):

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStoreHealthcheck.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    private final Path dataStorePath;
+
+    public FileSystemDataStoreHealthcheck(final Path dataStorePath)
+    {
+        this.dataStorePath = dataStorePath;
+    }
+
+    @Override
+    public HealthResult call() throws IOException
```
rorytorneymf, line 35: @dermot-hardy Can you re-review this file, its changed since you approved it, thanks.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -223,7 +237,25 @@ public void close() throws IOException
...
+        try {
+            return healthcheckFuture.get(healthcheckTimeout.getSeconds(), TimeUnit.SECONDS);
+        } catch (final TimeoutException e) {
+            healthcheckFuture.cancel(true);
+            return new HealthResult(
+                HealthStatus.UNHEALTHY,
+                String.format("Timeout after %s seconds trying to access data store directory %s",
+                              healthcheckTimeout.getSeconds(),
+                              dataStorePath.toString()));
+        } catch (final Exception e) {
```
dermot-hardy: You've changed this from `} catch (final InterruptedException | ExecutionException e) {` to `} catch (final Exception e) {`.  Are you trying to catch unchecked exceptions as well or why have you done this?

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -26,8 +27,16 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.*;
+import java.time.Duration;
 import java.util.Objects;
 import java.util.UUID;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
```
dermot-hardy, line 34: Unused import being added.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStoreHealthcheck.java
dermot-hardy: This file looks fine to me.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
dermot-hardy: Unless you had a reason for changing this I think you had it right before:
```suggestion
        } catch (final InterruptedException | ExecutionException e) {
```

dermot-hardy (approved):

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
rorytorneymf: Yes thanks, I mistakenly changed it to Exception so that it would catch the IOException thrown from the healthcheck as well, but I can see now that if an IOException is thrown, it will be wrapped in an ExecutionException. Reverted.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
rorytorneymf: Used again now.

dermot-hardy (approved):

PR #111, opened 2020-08-17
SCMOD-10661: Implement baseCancel for Queue Consumer

dgibson-microfocus:
https://portal.digitalsafe.net/browse/SCMOD-9988

Implements the `basicCancel` method for the RabbitMQ consumers. This is called in the event the broker itself has cancelled the consumer (usually due to the queue being deleted or the node owning the queue going down) which bypasses the `Channel`. See [RabbitMQ docs](https://www.rabbitmq.com/consumer-cancel.html)

### DIFF
--- a/util-rabbitmq/src/test/java/com/hpe/caf/util/rabbitmq/DefaultRabbitConsumerTest.java
+++ b/util-rabbitmq/src/test/java/com/hpe/caf/util/rabbitmq/DefaultRabbitConsumerTest.java
@@ -151,6 +151,12 @@ public class DefaultRabbitConsumerTest
             latch.countDown();
         }
 
+        @Override
+        public void processCancel(String consumerTag)
+        {
+            //no-op
+        }
+
         public Delivery getLastDelivery()
         {
             return lastDelivery;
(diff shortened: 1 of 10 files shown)


util-rabbitmq/src/test/java/com/hpe/caf/util/rabbitmq/DefaultRabbitConsumerTest.java
```diff
@@ -151,6 +151,12 @@ public void processDrop(long tag)
             latch.countDown();
         }
 
+        @Override
+        public void processCancel(String consumerTag)
+        {
```
DavidGerald-Milligan, line 156: Would this no-op be better as a default method on the interface?

DavidGerald-Milligan: Sorry I see this is a test

DavidGerald-Milligan (approved):

michael-bryson:
@dgibson-microfocus Is this PR still required or can it be closed?

dgibson-microfocus:
I don't think it's required

PR #125, opened 2021-03-15
SCMOD-12730: Handle paused jobs

rorytorneymf:
https://portal.digitalsafe.net/browse/SCMOD-12730

  - The `statusCheckUrl` workers use to check the status of a task is being changed to point to the `status` endpoint instead of the `isActive` endpoint: https://github.com/JobService/job-service/pull/237.
  - Instead of returning `true` or `false`, the `statusCheckUrl` will now return one of `Active`, `Cancelled`, `Completed`, `Failed`,
  `Paused`, or `Waiting`.
  - When a worker receives a task, it will now check if the task has been paused using the `statusCheckUrl`.
  - If the task has been paused, and the `CAF_WORKER_PAUSED_QUEUE` environment variable is set, the worker will publish the task to the `CAF_WORKER_PAUSED_QUEUE` instead of processing it.
  - If the task has been paused, and the `CAF_WORKER_PAUSED_QUEUE` environment variable is NOT set, the worker process the task as normal (as if the task was not paused).


worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -269,80 +327,88 @@ private boolean performJobStatusCheck(TaskMessage tm) throws InvalidJobTaskIdExc
...
+        private JobStatusResponse getJobStatus(String jobId, String statusCheckUrl) throws JobNotFoundException
         {
             JobStatusResponse jobStatusResponse = new JobStatusResponse();
             try {
                 URL url = new URL(statusCheckUrl);
-                URLConnection connection = url.openConnection();
+                final URLConnection connection = url.openConnection();
+                if (connection instanceof HttpURLConnection) {
+                    if (((HttpURLConnection) connection).getResponseCode() == HttpURLConnection.HTTP_NOT_FOUND) {
+                        throw new JobNotFoundException(
```
rorytorneymf, line 362: Previously, when a job was not found in the DB, the job service `isActive` endpoint would have returned a HTTP 200 status with a response body of `false`.

https://github.com/JobService/job-service/blob/develop/job-service/src/main/java/com/hpe/caf/services/job/api/DatabaseHelper.java#L469

This resulted in `WorkerCore` then discarding the message.

Because the `statusCheckUrl` now points to the `status` endpoint, which returns a status enum, I think the cleanest way to signify a job not found is to return a HTTP 404 status (unless we want to add a `NON_FOUND` value to the status enum?).

As such, we now need to check for a 404 here and discard the message if we get a 404:

https://github.com/WorkerFramework/worker-framework/pull/125/files#diff-b460c174826f9510c84f6a22b7c5471e5a81e20d5641b55774ae5dc1a0b75f20R167

thus maintaining the same behaviour as we previously had when a job was not found.

worker-queue-rabbit/readme.md
```diff
@@ -11,6 +11,7 @@
 
 - prefetchBuffer: the number of additional messages (tasks) to request from the RabbitMQ server beyond the number of tasks the worker can simultaneously handle. Minimum 0, Maximum 100000.
 - inputQueue: the routing key for a direct exchange (ie. queue name) to receive input tasks from, this must be set
+- pausedQueue: the routing key to use to send messages to when a worker is paused, this is optional, and if not set, messages sent to a paused worker will be processed as normal (as if the worker was not paused)
```
dermot-hardy: Reword - workers are never paused - jobs can be paused.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -269,80 +326,88 @@ private boolean performJobStatusCheck(TaskMessage tm) throws InvalidJobTaskIdExc
...
-                return true;
+                return JobStatus.Active;
             }
 
             String jobId = tracking.getJobId();
             LOG.debug("Task {} (job {}) - attempting to check job status", tm.getTaskId(), jobId);
-            JobStatusResponse jobStatus = getJobStatus(jobId, statusCheckUrl);
-            long newStatusCheckTime = System.currentTimeMillis() + jobStatus.getStatusCheckIntervalMillis();
+            JobStatusResponse jobStatusResponse = getJobStatus(jobId, statusCheckUrl);
+            long newStatusCheckTime = System.currentTimeMillis() + jobStatusResponse.getStatusCheckIntervalMillis();
```
dermot-hardy: Let's change the way the status check time is passed between the workers.  Instead of summing these and passing the sum let's pass them individually and allow the receiving worker to decide whether it should lookup the status or not.  For workers that have a paused queue I imagine they would want to lookup more frequently than the rest.

dermot-hardy (approved):

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
rorytorneymf: Added in this commit: https://github.com/WorkerFramework/worker-framework/pull/125/commits/07ddc671ae4404f751efe8497982d7ba171d6243

dermot-hardy (approved):

PR #126, opened 2021-04-07
SCMOD-12730: Allow task stowing worker to receive tasks

rorytorneymf:
https://portal.digitalsafe.net/browse/SCMOD-12730

### DIFF
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/NotIndendedTaskMessageForwardingEvaluator.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/NotIndendedTaskMessageForwardingEvaluator.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015-2021 Micro Focus or one of its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.api.worker;
+
+import com.hpe.caf.api.Codec;
+import java.util.Map;
+
+/**
+ * A worker should implement this interface if it is capable of evaluating task messages that are not intended for it (because the
+ * worker's input queue is different from the target/to queue in the task message), and deciding whether they are to be discarded,
+ * executed or forwarded.
+ */
+public interface NotIndendedTaskMessageForwardingEvaluator
+{
+    /**
+     * Examines the task message and decides whether to discard, execute or forward it.
+     *
+     * @param tm the task message
+     * @param taskInformation the reference to the message this task arrived on
+     * @param poison flag indicating if the message is a poison message
+     * @param headers the map of key/value paired headers to be stamped on the message
+     * @param codec the Codec that can be used to serialise/deserialise data
+     * @param jobStatus the job status as returned by the status check URL
+     * @param callback worker callback to enact the forwarding action determined by the worker
+     * @return the forwarding action to be taken by the worker
+     */
+    TaskForwardingAction determineForwardingAction(
+        TaskMessage tm,
+        TaskInformation taskInformation,
+        boolean poison,
+        Map<String, Object> headers,
+        Codec codec,
+        JobStatus jobStatus,
+        WorkerCallback callback);
+}
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/NotIntendedTaskMessageForwardingEvaluator.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/NotIntendedTaskMessageForwardingEvaluator.java
@@ -23,7 +23,7 @@ import java.util.Map;
  * worker's input queue is different from the target/to queue in the task message), and deciding whether they are to be discarded,
  * executed or forwarded.
  */
-public interface NotIndendedTaskMessageForwardingEvaluator
+public interface NotIntendedTaskMessageForwardingEvaluator
 {
     /**
      * Examines the task message and decides whether to discard, execute or forward it.
--- a/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
+++ b/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
@@ -97,14 +97,14 @@ final class WorkerExecutor
         }
         //Check whether this worker application can evaluate messages not indended for it for forwarding.
         else if (factory instanceof NotIndendedTaskMessageForwardingEvaluator) {
-            determineNotIntendedTaskForwardingAction(tm, taskInformation, poison, headers, codec, jobStatus);
+            processNotIntendedTaskForwardingAction(tm, taskInformation, poison, headers, codec, jobStatus);
         //Else messages are forwarded by default.
         } else {
             callback.forward(taskInformation, tm.getTo(), tm, headers);
         }
     }
 
-    private void determineNotIntendedTaskForwardingAction(final TaskMessage tm, final TaskInformation taskInformation,
+    private void processNotIntendedTaskForwardingAction(final TaskMessage tm, final TaskInformation taskInformation,
                                                           final boolean poison,
                                                           final Map<String, Object> headers, final Codec codec, final JobStatus jobStatus)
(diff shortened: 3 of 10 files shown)


worker-api/src/main/java/com/hpe/caf/api/worker/NotIndendedTaskMessageForwardingEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.hpe.caf.api.Codec;
+import java.util.Map;
+
+/**
+ * A worker should implement this interface if it is capable of evaluating task messages that are not intended for it (because the
+ * worker's input queue is different from the target/to queue in the task message), and deciding whether they are to be discarded,
+ * executed or forwarded.
+ */
+public interface NotIndendedTaskMessageForwardingEvaluator
```
rorytorneymf: I know we talked about having the new worker implement the existing `TaskMessageForwardingEvaluator` interface, but I couldn't see how to make that work for both the new worker, which does this:

https://github.com/JobService/worker-taskstowing/blob/SCMOD-12730/worker-taskstowing/src/main/java/com/microfocus/caf/worker/taskstowing/factory/TaskStowingWorkerFactory.java#L104

```
@Override
public TaskForwardingAction determineForwardingAction(
    final TaskMessage tm,
    final TaskInformation taskInformation,
    final boolean poison,
    final Map<String, Object> headers,
    final Codec codec,
    final JobStatus jobStatus,
    final WorkerCallback callback)
{
    return jobStatus == JobStatus.Paused
        ? TaskForwardingAction.Execute
        : TaskForwardingAction.Forward;
}
```
and the existing job tracking worker, which does this:

https://github.com/JobService/job-service/blob/develop/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerFactory.java

```
@Override
public void determineForwardingAction(TaskMessage proxiedTaskMessage, TaskInformation taskInformation,
                                      Map<String, Object> headers, WorkerCallback callback) {

    ...
    LOG.debug("Forwarding task {}", proxiedTaskMessage.getTaskId());
    callback.forward(taskInformation, proxiedTaskMessage.getTo(), proxiedTaskMessage, headers);
}
```
    
One possible way would be to update the existing interface to return `TaskForwardingAction` instead of `void`, but in order to maintain the current behavior of the job tracking worker, we would need to add a `NoOp` value to `TaskForwardingAction`, that the job tracking worker could return instead of `void`:

```
@Override
public TaskForwardingAction determineForwardingAction(TaskMessage proxiedTaskMessage, TaskInformation taskInformation, Map<String, Object> headers, WorkerCallback callback) {
    ...
    callback.forward(taskInformation, proxiedTaskMessage.getTo(), proxiedTaskMessage, headers);
    return TaskForwardingAction.NoOp;
}
```
    
Then the `WorkerExecutor` code might look something like this:

```
final TaskForwardingAction taskForwardingAction = ((TaskMessageForwardingEvaluator) factory).
            determineForwardingAction(tm, taskInformation, poison, headers, codec, jobStatus, callback);
switch (taskForwardingAction) {
    case Discard:
        discardTask(tm, taskInformation);
        break;
    case Execute:
        executeTask(tm, taskInformation, poison, headers, codec);
        break;
    case Forward:
        callback.forward(taskInformation, tm.getTo(), tm, headers);
        break;
    case NoOp:
        break;
    default:
        LOG.warn("Worker returned an unexpected forwarding action: {} for task: {} (message id: {}). "
            + "Defaulting to forwarding the task.", taskForwardingAction, tm.getTaskId(),
        callback.forward(taskInformation, tm.getTo(), tm, headers);
}
```

I'm not sure what other workers might be implementing `TaskMessageForwardingEvaluator` as well, so I'd need to check that and ensure they are updated as well if we go with this option.

What are your thoughts?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
```diff
@@ -78,23 +78,59 @@ public void executeTask(final TaskMessage tm, final TaskInformation taskInformat
...
      *
      * @param tm the task message
      * @param taskInformation the reference to the message this task arrived on
+     * @param poison flag indicating if the message is a poison message
      * @param headers the map of key/value paired headers to be stamped on the message
+     * @param codec the Codec that can be used to serialise/deserialise data
+     * @param jobStatus the job status as returned by the status check URL
      */
-    public void forwardTask(final TaskMessage tm, final TaskInformation taskInformation, Map<String, Object> headers) throws TaskRejectedException
+    public void forwardTask(final TaskMessage tm, final TaskInformation taskInformation, final boolean poison,
```
dermot-hardy: Given the change in functionality we should perhaps think about renaming this method - perhaps `handleDivertedTask`?

worker-api/src/main/java/com/hpe/caf/api/worker/NotIntendedTaskMessageForwardingEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.hpe.caf.api.Codec;
+import java.util.Map;
+
+/**
+ * A worker should implement this interface if it is capable of evaluating task messages that are not intended for it (because the
+ * worker's input queue is different from the target/to queue in the task message), and deciding whether they are to be discarded,
+ * executed or forwarded.
+ */
+public interface NotIntendedTaskMessageForwardingEvaluator
```
dermot-hardy: I know I'm making basically the same comment here again but I think we have the terminology wrong in this PR.  It's centered around forwarding even though that it only one of the 3 options you're making available.

Possibly consider renaming this to `DivertedTaskHandler`.

worker-api/src/main/java/com/hpe/caf/api/worker/NotIntendedTaskMessageForwardingEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.api.worker;
+
+import com.hpe.caf.api.Codec;
+import java.util.Map;
+
+/**
+ * A worker should implement this interface if it is capable of evaluating task messages that are not intended for it (because the
```
dermot-hardy: "that have been diverted" rather than "that are not intended for it"?

worker-api/src/main/java/com/hpe/caf/api/worker/NotIntendedTaskMessageForwardingEvaluator.java
```diff
@@ -0,0 +1,48 @@
...
+     * @param tm the task message
+     * @param taskInformation the reference to the message this task arrived on
+     * @param poison flag indicating if the message is a poison message
+     * @param headers the map of key/value paired headers to be stamped on the message
+     * @param codec the Codec that can be used to serialise/deserialise data
+     * @param jobStatus the job status as returned by the status check URL
+     * @param callback worker callback to enact the forwarding action determined by the worker
+     * @return the forwarding action to be taken by the worker
+     */
+    TaskForwardingAction determineForwardingAction(
```
dermot-hardy:
```suggestion
    DivertedTaskAction handleDivertedTask(
```

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerExecutor.java
```diff
@@ -78,23 +78,59 @@ public void executeTask(final TaskMessage tm, final TaskInformation taskInformat
...
+                            final Map<String, Object> headers, final Codec codec, final JobStatus jobStatus)
+        throws TaskRejectedException
     {
         //Check whether this worker application can evaluate messages for forwarding.
         if (factory instanceof TaskMessageForwardingEvaluator) {
             ((TaskMessageForwardingEvaluator) factory).determineForwardingAction(tm, taskInformation, headers, callback);
+        }
+        //Check whether this worker application can evaluate messages not indended for it for forwarding.
+        else if (factory instanceof NotIntendedTaskMessageForwardingEvaluator) {
```
dermot-hardy: This new interface feels like it could entirely replace `TaskMessageForwardingEvaluator`.  I'd be tempted to have it take priority (i.e. check for it first), and to deprecate `TaskMessageForwardingEvaluator` in favor of it (i.e. add the `@Deprecated` annotation to the `TaskMessageForwardingEvaluator` interface).

dermot-hardy (changes requested):
Only terminology comments.

worker-api/src/main/java/com/hpe/caf/api/worker/NotIndendedTaskMessageForwardingEvaluator.java
dermot-hardy: What you did in supporting a new interface and deprecating the old one is fine.

dermot-hardy (approved):

PR #128, opened 2021-05-05
SCMOD-13301 Worker-Framework: Move away from Files.exists

simranjeetc:
https://portal.digitalsafe.net/browse/SCMOD-13301

### DIFF
--- a/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
+++ b/worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
@@ -65,12 +65,11 @@ public class FileSystemDataStore implements ManagedDataStore, FilePathProvider, 
         throws DataStoreException
     {
         dataStorePath = FileSystems.getDefault().getPath(config.getDataDir());
-        if (!doesFileExist(dataStorePath)) {
-            try {
-                Files.createDirectory(dataStorePath);
-            } catch (IOException e) {
-                throw new DataStoreException("Cannot create data store directory", e);
-            }
+        throwExceptionIfPathDoesNotExist(dataStorePath);
+        try {
+            Files.createDirectory(dataStorePath);
+        } catch (IOException e) {
+            throw new DataStoreException("Cannot create data store directory", e);
         }
         healthcheck = new FileSystemDataStoreHealthcheck(dataStorePath);
         healthcheckTimeout = getDataDirHealthcheckTimeoutSeconds(config);
@@ -80,25 +79,13 @@ public class FileSystemDataStore implements ManagedDataStore, FilePathProvider, 
         LOG.debug("Initialised");
     }
     
-    private boolean doesFileExist(final Path path)
+    private void throwExceptionIfPathDoesNotExist(final Path path) throws ReferenceNotFoundException
     {
-        try
-        {
+        try {
             path.getFileSystem().provider().checkAccess(path);
+        } catch(final IOException ex) {
+            throw new ReferenceNotFoundException("Reference not found ", ex);
         }
-        catch(final IOException ex)
-        {
-            if(ex instanceof NoSuchFileException)
-            {
-                LOG.error("file does not exist", ex);
-            }
-            else
-            {
-                LOG.error("File existence check failed ", ex);
-            }
-            return false;
-        }
-        return true;
     }
     
     @Override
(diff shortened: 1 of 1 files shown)


worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -79,7 +79,28 @@ public FileSystemDataStore(final FileSystemDataStoreConfiguration config)
 
         LOG.debug("Initialised");
     }
-
+    
+    private boolean doesFileExist(final Path path)
```
dermot-hardy: Utility functions should be `static`.

dermot-hardy (commented):
What you've done here is really debugging code for [SCMOD-13041](https://portal.digitalsafe.net/browse/SCMOD-13041) rather than the replacement implementation expected for [SCMOD-13301](https://portal.digitalsafe.net/browse/SCMOD-13301).  We can go ahead and check this in but make sure it is marked up as progressing SCMOD-13041.  For SCMOD-13301 the exception needs to be propagated so that the calling code can decide what to do.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -79,7 +79,28 @@ public FileSystemDataStore(final FileSystemDataStoreConfiguration config)
...
+                LOG.error("file does not exist", ex);
+            }
+            else
+            {
+                LOG.error("File existence check failed ", ex);
+            }
+            return false;
+        }
+        return true;
+    }
```
dermot-hardy: As as aside, this function would normally be formatted like this:

    private boolean doesFileExist(final Path path)
    {
        try {
            path.getFileSystem().provider().checkAccess(path);
        } catch (final IOException ex) {
            if (ex instanceof NoSuchFileException) {
                LOG.error("file does not exist", ex);
            } else {
                LOG.error("File existence check failed ", ex);
            }
            return false;
        }
        return true;
    }

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -65,11 +65,12 @@ public FileSystemDataStore(final FileSystemDataStoreConfiguration config)
         throws DataStoreException
     {
         dataStorePath = FileSystems.getDefault().getPath(config.getDataDir());
-        throwExceptionIfPathDoesNotExist(dataStorePath);
-        try {
-            Files.createDirectory(dataStorePath);
-        } catch (IOException e) {
-            throw new DataStoreException("Cannot create data store directory", e);
+        if (!Files.exists(dataStorePath)) {
```
simranjeetc: @dermot-hardy Do we need to replace this as well? Here non-existence is expected whereas in places where we are replacing Files.exists, existence of path is expected. If we still want to replace this, then we may need to gobble up the exception and just log it. 

Let me know what you think.

dermot-hardy (changes requested):
All usages of `Files.exists()` should be replaced.  This Jira is effectively banning the use of this method from our codebase because of its issues.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
```diff
@@ -79,7 +79,27 @@ public FileSystemDataStore(final FileSystemDataStoreConfiguration config)
 
         LOG.debug("Initialised");
     }
-
+    
+    private static boolean pathDoesNotExist(final Path path)
```
dermot-hardy: I must not be explaining this Jira correctly.  You've just replaced the `Files.exists` call with a method which has exactly the same problem.

dermot-hardy (changes requested):

dermot-hardy:
This is what a `doesFileExist()` method should look like:

    public static boolean doesFileExist(final Path path) throws IOException
    {
        try {
            path.getFileSystem().provider().checkAccess(path);
        } catch (final NoSuchFileException ex) {
            return false;
        }

        return true;
    }

Note that it only returns that the file does not exist if the exception when checking the access was `NoSuchFileException`.  Other `IOException`s are propagated.

worker-store-fs/src/main/java/com/hpe/caf/worker/datastore/fs/FileSystemDataStore.java
simranjeetc: Resolved with [this](https://github.com/WorkerFramework/worker-framework/pull/128#issuecomment-840929683) implementation.

dermot-hardy (approved):

PR #135, opened 2021-07-08
SCMOD-5524 Worker Framework: Eliminate base64 encoding of taskData to reduce message size

simranjeetc:
https://portal.digitalsafe.net/browse/SCMOD-5524

With this change, the worker would be able to publish messages in newer format only if it received one with that format. Otherwise, it will continue to fallback to sending messages in older format.

### DIFF
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
@@ -29,7 +29,7 @@ public interface WorkerCallback
      * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
      *                      taskData as object or byte[].
      */
-    void send(TaskInformation taskInformation, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
+    void send(TaskInformation taskInformation, TaskMessage responseMessage, boolean publishTaskDataAsObject);
 
     /**
      * Indicate a task was completed.
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
@@ -50,13 +50,18 @@ public interface WorkerCallback
 
     /**
      * Indicates the Worker wishes to forward this task to the specified queue without processing it.
-     *
-     * @param taskInformation a queue-specific reference for the incoming message to be forwarded
+     *  @param taskInformation a queue-specific reference for the incoming message to be forwarded
      * @param queue the queue to hold the forwarded message
      * @param forwardedMessage the message to put on the queue
      * @param headers the map of key/value paired headers to be stamped on the message
+     * @param sendNewFormat boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void forward(TaskInformation taskInformation, String queue, TaskMessage forwardedMessage, Map<String, Object> headers);
+    void forward(TaskInformation taskInformation,
+                 String queue,
+                 TaskMessage forwardedMessage,
+                 Map<String,Object> headers,
+                 boolean sendNewFormat);
 
     /**
      * Indicates the Worker wishes to forward this task to the specified paused queue without processing it.
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
@@ -24,20 +24,20 @@ public interface WorkerCallback
 {
     /**
      * Used to send responses prior to the final response (when complete should be used instead).
-     *
-     * @param taskInformation a queue-specific reference for the incoming message that generated the response
+     *  @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param responseMessage the message to put on the queue (target specified by the {@code to} property)
+     * @param sendNewFormat
      */
-    void send(TaskInformation taskInformation, TaskMessage responseMessage);
+    void send(TaskInformation taskInformation, TaskMessage responseMessage, final boolean sendNewFormat);
 
     /**
      * Indicate a task was completed.
-     *
-     * @param taskInformation a queue-specific reference for the incoming message that generated the response
+     *  @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param queue the queue to hold the message
      * @param responseMessage the message to put on the queue
+     * @param sendNewFormat
      */
-    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage);
+    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage, final boolean sendNewFormat);
 
     /**
      * Indicates the Worker wishes to abandon this task, but return it to the queue so that it can be retried by this or another Worker
@@ -77,9 +77,9 @@ public interface WorkerCallback
 
     /**
      * Used to send a report update message.
-     *
-     * @param taskInformation a queue-specific reference for the incoming message
+     *  @param taskInformation a queue-specific reference for the incoming message
      * @param reportUpdateMessage the report update message to put on the queue
+     * @param sendNewFormat
      */
-    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage);
+    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage, final boolean sendNewFormat);
 }
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
@@ -26,18 +26,20 @@ public interface WorkerCallback
      * Used to send responses prior to the final response (when complete should be used instead).
      *  @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param responseMessage the message to put on the queue (target specified by the {@code to} property)
-     * @param sendNewFormat
+     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void send(TaskInformation taskInformation, TaskMessage responseMessage, final boolean sendNewFormat);
+    void send(TaskInformation taskInformation, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
 
     /**
      * Indicate a task was completed.
      *  @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param queue the queue to hold the message
      * @param responseMessage the message to put on the queue
-     * @param sendNewFormat
+     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage, final boolean sendNewFormat);
+    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
 
     /**
      * Indicates the Worker wishes to abandon this task, but return it to the queue so that it can be retried by this or another Worker
@@ -54,14 +56,14 @@ public interface WorkerCallback
      * @param queue the queue to hold the forwarded message
      * @param forwardedMessage the message to put on the queue
      * @param headers the map of key/value paired headers to be stamped on the message
-     * @param sendNewFormat boolean flag which decides whether the message will be published with
+     * @param publishTaskDataAsObj boolean flag which decides whether the message will be published with
      *                      taskData as object or byte[].
      */
     void forward(TaskInformation taskInformation,
                  String queue,
                  TaskMessage forwardedMessage,
                  Map<String,Object> headers,
-                 boolean sendNewFormat);
+                 boolean publishTaskDataAsObj);
 
     /**
      * Indicates the Worker wishes to forward this task to the specified paused queue without processing it.
@@ -84,7 +86,8 @@ public interface WorkerCallback
      * Used to send a report update message.
      *  @param taskInformation a queue-specific reference for the incoming message
      * @param reportUpdateMessage the report update message to put on the queue
-     * @param sendNewFormat
+     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage, final boolean sendNewFormat);
+    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage, final boolean publishTaskDataAsObject);
 }
--- a/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
+++ b/worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
@@ -27,10 +27,8 @@ public interface WorkerCallback
      *
      * @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param responseMessage the message to put on the queue (target specified by the {@code to} property)
-     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
-     *                      taskData as object or byte[].
      */
-    void send(TaskInformation taskInformation, TaskMessage responseMessage, boolean publishTaskDataAsObject);
+    void send(TaskInformation taskInformation, TaskMessage responseMessage);
 
     /**
      * Indicate a task was completed.
@@ -38,10 +36,8 @@ public interface WorkerCallback
      * @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param queue the queue to hold the message
      * @param responseMessage the message to put on the queue
-     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
-     *                      taskData as object or byte[].
      */
-    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
+    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage);
 
     /**
      * Indicates the Worker wishes to abandon this task, but return it to the queue so that it can be retried by this or another Worker
@@ -59,14 +55,8 @@ public interface WorkerCallback
      * @param queue the queue to hold the forwarded message
      * @param forwardedMessage the message to put on the queue
      * @param headers the map of key/value paired headers to be stamped on the message
-     * @param publishTaskDataAsObj boolean flag which decides whether the message will be published with
-     *                      taskData as object or byte[].
      */
-    void forward(TaskInformation taskInformation,
-                 String queue,
-                 TaskMessage forwardedMessage,
-                 Map<String, Object> headers,
-                 boolean publishTaskDataAsObj);
+    void forward(TaskInformation taskInformation, String queue, TaskMessage forwardedMessage, Map<String, Object> headers);
 
     /**
      * Indicates the Worker wishes to forward this task to the specified paused queue without processing it.
@@ -90,8 +80,6 @@ public interface WorkerCallback
      *
      * @param taskInformation a queue-specific reference for the incoming message
      * @param reportUpdateMessage the report update message to put on the queue
-     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
-     *                      taskData as object or byte[].
      */
-    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage, final boolean publishTaskDataAsObject);
+    void reportUpdate(final TaskInformation taskInformation, final TaskMessage reportUpdateMessage);
 }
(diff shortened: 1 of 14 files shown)


worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -24,20 +24,22 @@
 {
     /**
      * Used to send responses prior to the final response (when complete should be used instead).
-     *
-     * @param taskInformation a queue-specific reference for the incoming message that generated the response
+     *  @param taskInformation a queue-specific reference for the incoming message that generated the response
```
dermot-hardy: JavaDoc formatting screwed up in this file.  It is correct for there to be a blank line before the parameters and only one space before `@param`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java
```diff
@@ -72,17 +73,18 @@ public WorkerTaskImpl(
             final TaskInformation taskInformation,
             final TaskMessage taskMessage,
             final boolean poison,
-            final Map<String, Object> headers,
+            final Map<String,Object> headers,
```
dermot-hardy: Why is this line changing?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -156,9 +156,14 @@ public void registerNewTask(final TaskInformation taskInformation, final byte[]
         private void registerNewTaskImpl(final TaskInformation taskInformation, final byte[] taskMessage, Map<String, Object> headers)
             throws InvalidTaskException, TaskRejectedException
         {
+            boolean publishTaskDataAsObject = false;
```
dermot-hardy: If you're not using `final` you should be asking yourself if there's a way you can adjust the code so that you can.  In this case I don't see why this line can't be removed and that `if` block just below be replaced with:

    final boolean publishTaskDataAsObject = !QueueTaskMessageFunctions.isTaskDataString(queueTaskMessage);

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -24,20 +24,22 @@
...
      * @param responseMessage the message to put on the queue (target specified by the {@code to} property)
+     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void send(TaskInformation taskInformation, TaskMessage responseMessage);
+    void send(TaskInformation taskInformation, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
```
dermot-hardy: Interface parameters should not be marked `final` (unless it is a default method)
```suggestion
    void send(TaskInformation taskInformation, TaskMessage responseMessage, boolean publishTaskDataAsObject);
```

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -50,13 +52,18 @@
...
-     * @param taskInformation a queue-specific reference for the incoming message to be forwarded
+     *  @param taskInformation a queue-specific reference for the incoming message to be forwarded
      * @param queue the queue to hold the forwarded message
      * @param forwardedMessage the message to put on the queue
      * @param headers the map of key/value paired headers to be stamped on the message
+     * @param publishTaskDataAsObj boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void forward(TaskInformation taskInformation, String queue, TaskMessage forwardedMessage, Map<String, Object> headers);
+    void forward(TaskInformation taskInformation,
```
dermot-hardy: Does `pause()` need updated as well?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -591,27 +598,32 @@ public void discard(TaskInformation taskInformation)
...
+                final int priority = reportUpdateMessage.getPriority() == null ? 0 : reportUpdateMessage.getPriority();
+        
                 workerQueue.publish(taskInformation, output, reportUpdateMessage.getTo(), Collections.emptyMap(), priority);
-            } catch (final QueueException ex) {
+            } catch (final QueueException | CodecException ex) {
                 throw new RuntimeException(ex);
             }
         }
+    
+        private byte[] convertAndSerializeMessage(final TaskMessage responseMessage, final boolean publishTaskDataAsObject) throws CodecException
```
dermot-hardy: Line goes over the 138-character guideline.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -451,27 +456,22 @@ public CoreWorkerCallback(final Codec codec, final ManagedWorkerQueue workerQueu
...
-            }
-
-            final int priority = responseMessage.getPriority() == null ? 0 : responseMessage.getPriority();
-
             try {
+    
+                final byte[] output = convertAndSerializeMessage(responseMessage, publishTaskDataAsObject);
+    
+                final int priority = responseMessage.getPriority() == null ? 0 : responseMessage.getPriority();
+
```
dermot-hardy: Remove the whitespace from the lines.

```suggestion
                final byte[] output = convertAndSerializeMessage(responseMessage, publishTaskDataAsObject);

                final int priority = responseMessage.getPriority() == null ? 0 : responseMessage.getPriority();

```

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -505,11 +508,11 @@ public void complete(final TaskInformation taskInformation, final String queue,
                     // messages.
                     workerQueue.acknowledgeTask(taskInformation);
                 } else {
-                    // **** Normal Worker ****                    
+                    // **** Normal Worker ****
```
dermot-hardy: I agree with the change but it's not part of this Jira.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -505,11 +508,11 @@ public void complete(final TaskInformation taskInformation, final String queue,
...
                     // A worker with an input and output queue.
-                    byte[] output = codec.serialise(responseMessage);
-                    workerQueue.publish(taskInformation, output, queue, Collections.emptyMap(), 
-                                           responseMessage.getPriority() == null ? 0 : responseMessage.getPriority(), true);                    
+                    final byte[] output = convertAndSerializeMessage(responseMessage, publishTaskDataAsObject);
+                    workerQueue.publish(taskInformation, output, queue, Collections.emptyMap(),
+                                           responseMessage.getPriority() == null ? 0 : responseMessage.getPriority(), true);
```
dermot-hardy: Again I agree with the change but it's making me spend time determining what has changed only to discover it is just whitespace.  Keep your own whitespace tidy in lines you're introducing or changing but don't go fixing it unnecessarily in lines you're not otherwise changing.  There is no issue having a separate commit for general whitespace tidy-ups.

```suggestion
                    workerQueue.publish(taskInformation, output, queue, Collections.emptyMap(), 
                                           responseMessage.getPriority() == null ? 0 : responseMessage.getPriority(), true);                    
```

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -549,7 +556,7 @@ public void forward(TaskInformation taskInformation, String queue, TaskMessage f
                     workerQueue.acknowledgeTask(taskInformation);
                 } else {
                     // Else forward the task
-                    byte[] output = codec.serialise(forwardedMessage);
+                    final byte[] output = convertAndSerializeMessage(forwardedMessage, publishTaskDataAsObj);
```
dermot-hardy: Why not `pause()` also?

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -591,27 +598,32 @@ public void discard(TaskInformation taskInformation)
...
-                throw new RuntimeException(ex);
-            }
-
-            final int priority = reportUpdateMessage.getPriority() == null ? 0 : reportUpdateMessage.getPriority();
-
-            try {                
+                final byte[] output = convertAndSerializeMessage(reportUpdateMessage, publishTaskDataAsObject);
+        
+                final int priority = reportUpdateMessage.getPriority() == null ? 0 : reportUpdateMessage.getPriority();
+
```
dermot-hardy:
```suggestion

                final int priority = reportUpdateMessage.getPriority() == null ? 0 : reportUpdateMessage.getPriority();

```

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -591,27 +598,32 @@ public void discard(TaskInformation taskInformation)
...
+        
+                final int priority = reportUpdateMessage.getPriority() == null ? 0 : reportUpdateMessage.getPriority();
+        
                 workerQueue.publish(taskInformation, output, reportUpdateMessage.getTo(), Collections.emptyMap(), priority);
-            } catch (final QueueException ex) {
+            } catch (final QueueException | CodecException ex) {
                 throw new RuntimeException(ex);
             }
         }
+
```
dermot-hardy:
```suggestion

```

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
```diff
@@ -24,20 +24,22 @@
...
-     *
-     * @param taskInformation a queue-specific reference for the incoming message that generated the response
+     *  @param taskInformation a queue-specific reference for the incoming message that generated the response
      * @param queue the queue to hold the message
      * @param responseMessage the message to put on the queue
+     * @param publishTaskDataAsObject boolean flag which decides whether the message will be published with
+     *                      taskData as object or byte[].
      */
-    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage);
+    void complete(TaskInformation taskInformation, String queue, TaskMessage responseMessage, final boolean publishTaskDataAsObject);
```
dermot-hardy: Looking at it more I see that this interface is exposed by [`TaskMessageForwardingEvaluator`](https://github.com/WorkerFramework/worker-framework/blob/v5.0.0/worker-api/src/main/java/com/hpe/caf/api/worker/TaskMessageForwardingEvaluator.java#L38) and [`DivertedTaskHandler`](https://github.com/WorkerFramework/worker-framework/blob/v5.0.0/worker-api/src/main/java/com/hpe/caf/api/worker/DivertedTaskHandler.java#L46).  Unfortunately that means that adding these parameters is a breaking change.

Here is some code in Job Service which wouldn't compile after these changes for example:
[JobTrackingWorkerFactory.java:272](https://github.com/JobService/job-service/blob/v4.0.0-source/worker-jobtracking/src/main/java/com/hpe/caf/worker/jobtracking/JobTrackingWorkerFactory.java#L272)

Possible fixes:
1. Accept it - move to `6.0.0-SNAPSHOT` and declare the break.
2. Add default methods to the interface with the old signatures.  They could just call the new methods and pass `false` for the `publishTaskDataAsObject` parameter.
3. Revert the changes to the interface and pass the information in a different way (such as using [`ThreadLocal`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) or [`WeakHashMap`](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) for example.

It may seem hacky but in this case I'm in favor of option 3.  Using `ThreadLocal` wouldn't work as there is no guarantee that the callback would be on the same thread (in fact you're guaranteed that it won't be).  But a `WeakHashMap` could be used to tag the extra detail to the `TaskInformation` object.

Create the map in the `WorkerCore` constructor and then pass it to both the `CoreWorkerCallback` and `CoreTaskCallback` classes:

    final Map<TaskInformation, QueueTaskMessage> originalMessageLookup = Collections.synchronizedMap(new WeakHashMap<>());

The `CoreTaskCallback` code can populate the map and the `CoreWorkerCallback` code could make use of it.

I think taking this approach should also eliminate a lot of the plumbing changes in this PR (to `WorkerExecutor` and `WorkerTaskImpl` for example).

dermot-hardy (changes requested):

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Whitespace corrections have creeped in due to auto-format. I'll see how I can correct that in IDE configuration.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
dermot-hardy: Other options:
- Use Guava's or another alternative to `WeakHashMap` - similar but potentially better concurrency
- Create a `CoreTaskInformation` object which implements `TaskInformation`, storing both the underlying `TaskInformation` object, and the `QueueTaskMessage` object.  These would need to be created in `CoreTaskCallback` but could then be unwrapped in `CoreWorkerCallback` to get the original objects back.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Reverted at db47d48

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Reverted at db47d48

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
simranjeetc: Resolved here and at other places in this file.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Reverted at db47d48

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Ideally, it is required for pause(or anything else which sends message to the pipes). I had checked with @rorytorneymf though and found that pause/unpause is on hold at the moment. So, thought of not making any changes since a corresponding change is required for unpause which is not merged yet. 
But now it seems harmless even if the change is done for pause. So will take care of it while implementing one of the other approaches brought out by you for handling this change.

worker-api/src/main/java/com/hpe/caf/api/worker/WorkerCallback.java
simranjeetc: Used

 > Create a `CoreTaskInformation` object which implements `TaskInformation`, storing both the underlying `TaskInformation` object, and the `QueueTaskMessage` object.  These would need to be created in `CoreTaskCallback` but could then be unwrapped in `CoreWorkerCallback` to get the original objects back.

in 8c8dafd

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: Addressed this for pause as well in 8c8dafd. Resolving other comments where same thing is pointed out.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: > Again I agree with the change but it's making me spend time determining what has changed only to discover it is just whitespace. Keep your own whitespace tidy in lines you're introducing or changing but don't go fixing it unnecessarily in lines you're not otherwise changing. There is no issue having a separate commit for general whitespace tidy-ups.

Fixed the whitespace. But you may see additional whitespace changes where I am changing the parameters to be on separate lines. This is being done in places where after my change the character count is spilling over 138.

worker-core/src/main/java/com/hpe/caf/worker/core/CoreTaskInformation.java
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.hpe.caf.worker.core;
+
+import com.hpe.caf.api.worker.QueueTaskMessage;
+import com.hpe.caf.api.worker.TaskInformation;
+
+public class CoreTaskInformation implements TaskInformation
```
dermot-hardy: Does this need to be `public`?  Couldn't you get away with package-private as with `WorkerCore` itself?

worker-core/src/main/java/com/hpe/caf/worker/core/CoreTaskInformation.java
```diff
@@ -0,0 +1,51 @@
...
+{
+    private final String inboundMessageId;
+    private final TaskInformation taskInformation;
+    private final QueueTaskMessage queueTaskMessage;
+
+    public CoreTaskInformation(final String inboundMessageId,
```
dermot-hardy: I don't think you need to take `inboundMessageId` as a parameter.  You don't even really need to store it since the `getInboundMessageId()` implementation could be `return taskInformation.getInboundMessageId();`.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -507,8 +504,8 @@ public void complete(final TaskInformation taskInformation, final String queue,
                 } else {
                     // **** Normal Worker ****                    
                     // A worker with an input and output queue.
-                    byte[] output = codec.serialise(responseMessage);
-                    workerQueue.publish(taskInformation, output, queue, Collections.emptyMap(), 
+                    final byte[] output = convertAndSerializeMessage(responseMessage, taskInformation);
+                    workerQueue.publish(getTaskInformation(taskInformation), output, queue, Collections.emptyMap(),
```
dermot-hardy: The unwrapping code is too intermingled here.  For example we are unwrapping here for calling `workerQueue.publish` but not unwrapping just above for `workerQueue.acknowledgeTask`, or for the `abandon` code.  It needs to be centralised.

dermot-hardy (changes requested):

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
dermot-hardy: I wasn't sure that this comment was clear to I've applied it myself: 486ea3e.
The unwrapping is now done in a separate layer introduced for the purpose.  This will be easier for future developers to maintain.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -172,33 +175,33 @@ private void registerNewTaskImpl(final TaskInformation taskInformation, final by
...
                                     taskInformation.getInboundMessageId());
-                                executor.executeTask(tm, taskInformation, poison, headers, codec);
+                                executor.executeTask(tm, coreTaskInformation, poison, headers, codec);
                             }
                         } else {
-                            executor.handleDivertedTask(tm, taskInformation, poison, headers, codec, jobStatus);
+                            executor.handleDivertedTask(tm, coreTaskInformation, poison, headers, codec, jobStatus);
                         }
                         break;
                     default:
```
dermot-hardy, line 205: `executor.discardTask` not passing wrapped task information.

dermot-hardy (changes requested):

worker-core/src/main/java/com/hpe/caf/worker/core/CoreTaskInformation.java
dermot-hardy: Fixed in dafbbbe.

worker-core/src/main/java/com/hpe/caf/worker/core/CoreTaskInformation.java
dermot-hardy: Fixed in 59e800c.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
dermot-hardy: Fixed in 4171002.

dermot-hardy (approved):

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
simranjeetc: `discardTask` required only the `inboundMessageId` which was available from the `taskInformation` too. Isn't wrapping required only when we need access to `taskData` ? Or is this being done to maintain consistency and not maintain two versions of the similar object?

dermot-hardy: I figured that might be why you left it out and you're right that with the current implementation that it wouldn't cause an issue.  But the code overall is easier to reason about if the `WorkerExecutor` is only ever passed the wrapped task information.  You always know it has to be unwrapped in the callback.  (Looking at the changes now I suspect that the `WeakHashMap` solution would probably have been less changes - but I also think this solution is fine).

PR #143, opened 2022-12-12
572084: Update jackson to 2.13.4

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=572084


dermot-hardy:
We're now up to Jackson 2.15.3.

PR #144, opened 2023-01-12
US614206: Worker Framework DivertedTask Changes

andyreidz:
Added a new Env CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING set this to false to disable diverted task checking

### DIFF
--- a/release-notes-6.0.0.md
+++ b/release-notes-6.0.0.md
@@ -4,6 +4,8 @@
 ${version-number}
 
 #### New Features
+- 614206: Diverted message checking can now be disabled using the CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING environment 
+variable.
 
 #### Breaking Changes
 - 378627: Removes worker-testing-util and worker-testing-integration
--- a/worker-core/readme.md
+++ b/worker-core/readme.md
@@ -76,6 +76,9 @@
 
  The following environment variables can optionally be set:
  - caf.appname: manual service name to be set if not in a Marathon environment
+ - CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING: A boolean that indicates if the `to` field of a TaskMessage should be 
+compared to the input current of the worker, redirecting the message to the queue in the `to` field if it does not match 
+the current input queue. Default is True.
 
 ### Starting the application
 
--- a/worker-core/readme.md
+++ b/worker-core/readme.md
@@ -77,7 +77,7 @@
  The following environment variables can optionally be set:
  - caf.appname: manual service name to be set if not in a Marathon environment
  - CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING: A boolean that indicates if the `to` field of a TaskMessage should be 
-compared to the input current of the worker, redirecting the message to the queue in the `to` field if it does not match 
+compared to the input queue of the worker, redirecting the message to the queue in the `to` field if it does not match 
 the current input queue. Default is True.
 
 ### Starting the application
--- a/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
+++ b/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
@@ -48,6 +48,8 @@ final class WorkerCore
     private final TaskCallback callback;
     private static final Logger LOG = LoggerFactory.getLogger(WorkerCore.class);
     private boolean isStarted;
+    private static final boolean isDivertedTaskCheckingEnabled = Boolean.parseBoolean(
+        System.getProperty("CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING", "true"));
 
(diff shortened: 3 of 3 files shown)


worker-core/readme.md
```diff
@@ -76,6 +76,9 @@
 
  The following environment variables can optionally be set:
  - caf.appname: manual service name to be set if not in a Marathon environment
+ - CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING: A boolean that indicates if the `to` field of a TaskMessage should be 
+compared to the input current of the worker, redirecting the message to the queue in the `to` field if it does not match
```
rorytorneymf:
```suggestion
compared to the input queue of the worker, redirecting the message to the queue in the `to` field if it does not match 
```

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -252,6 +252,10 @@ private void validateTaskMessage(TaskMessage tm) throws InvalidTaskException
 
         private boolean isTaskIntendedForThisWorker(final TaskMessage tm, final TaskInformation taskInformation)
         {
+            if (!Boolean.parseBoolean(
```
rorytorneymf: (Optional) Is it worth adding a debug log here?

```
final boolean divertedTaskCheckingEnabled
    = Boolean.parseBoolean(System.getProperty("CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING", "true"));

if (!divertedTaskCheckingEnabled) {
    LOG.debug(
        "Diverted task checking is disabled, so assuming that Task {} (message id: {}) on input queue {} "
        + "is intended for this worker",
        tm.getTaskId(),
        taskInformation.getInboundMessageId(),
        workerQueue.getInputQueue());

    return true;
}
```

rorytorneymf (approved):

rorytorneymf (approved):

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -252,6 +252,18 @@ private void validateTaskMessage(TaskMessage tm) throws InvalidTaskException
...
+                    System.getProperty("CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING", "true"))) {
```
dermot-hardy: Do a `private static final` so that this isn't re-parsed for every message.

Like this one: [WorkerTaskImpl.java:49-50](https://github.com/WorkerFramework/worker-framework/blob/v5.1.0/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerTaskImpl.java#L49-L50)

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerCore.java
```diff
@@ -252,6 +252,18 @@ private void validateTaskMessage(TaskMessage tm) throws InvalidTaskException
...
+
+                LOG.debug(
+                    "Diverted task checking is disabled, so assuming that Task {} (message id: {}) on input queue {} "
+                            + "is intended for this worker",
+                    tm.getTaskId(),
+                    taskInformation.getInboundMessageId(),
+                    workerQueue.getInputQueue());                
+
```
dermot-hardy: Remove the extra blank lines at the end of the line:

```suggestion
                    workerQueue.getInputQueue());

```

dermot-hardy (approved):

release-notes-6.0.0.md
```diff
@@ -4,6 +4,8 @@
 ${version-number}
 
 #### New Features
+- 614206: Diverted message checking can now be disabled using the CAF_WORKER_ENABLE_DIVERTED_TASK_CHECKING environment 
+variable.
```
rorytorneymf, line 8: @andyreidz / @caoimhelynch Should duplicate this info on the new env variable in the [worker-document](https://github.com/CAFDataProcessing/worker-document/blob/develop/release-notes-5.2.0.md) release notes as well?

For example, the App Resources Worker pulls in the latest Worker Document version in order to get this new env var functionality from the Worker Framework:

https://github.houston.softwaregrp.net/Verity/darwin-worker-appresources/pull/106/files

andyreidz: I'm ok with not duplicating it, they can find what they need by following through the documentation of document worker and worker framework.

rorytorneymf: Thanks, looking at the [worker-document commits](https://github.com/CAFDataProcessing/worker-document/commits/develop), this is the only meaningful change though, and it is a new feature, so if we don't put it in, we have nothing to put in the worker-document release notes.

Normally, if there are only dependency updates, we change the release to a patch release and just have this as the release notes:

```
#### Version Number
${version-number}

#### New Features
- None

#### Bug Fixes
- Dependency updates only in this release.

#### Known Issues
- None
``` 

However, I'm not sure if this is right for the worker-document given the change above (i.e. the new env var is a new feature), so maybe we should include it in the worker-document release notes?

dermot-hardy: There should be an entry under New Features as you say, to justify the minor version bump, but it could just say "Upgraded to new version of Worker Framework" or something like that if that makes more sense than repeating details.

PR #153, opened 2023-07-05
743080: Excessive calls to queueDeclare

andyreidz:
The Set<> that was used to record queues that had previously been declared was not being populated.
Queues should never be deleted while workers are running, because that would risk losing a message, so this simple cache is acceptable.

### DIFF
--- a/release-notes-6.1.0.md
+++ b/release-notes-6.1.0.md
@@ -5,4 +5,7 @@ ${version-number}
 
 #### New Features
 
+#### Bug Fixes
+- 743080: Excessive calls to queueDeclare have been prevented by recording previously declared queues.
+
 #### Known Issues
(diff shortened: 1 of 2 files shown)


frankmccarry (approved):

andyreidz:
@dermot-hardy Just wanted to check that 6.1.0 is an acceptable version number for this change?

dermot-hardy:
> 6.1.0 is an acceptable version number for this change?

Yes but it would need to be written up as a new feature in the release notes rather than as a bug fix.

michael-bryson:
> > 6.1.0 is an acceptable version number for this change?
> 
> Yes but it would need to be written up as a new feature in the release notes rather than as a bug fix.

@andyreidz Looks like this comment was not addressed, the change is still listed as a Bug Fix in the release notes. If we are moving to 6.1.0 this needs to be listed as a New Feature, otherwise we can re-version the project to 6.0.1.

PR #154, opened 2023-10-02
US807003: Add a test worker and check that it starts

dermot-hardy:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=807003


michael-bryson (approved):
LGTM

PR #161, opened 2023-11-01
854021: Revert changes made in SCMOD-5524 - Eliminate base64 encoding of taskData to reduce message size

andyreidz:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=854021


dermot-hardy (approved):

worker-api/src/main/java/com/hpe/caf/api/worker/QueueTaskMessage.java
```diff
@@ -1,261 +0,0 @@
...
-import java.util.Objects;
-
-import jakarta.validation.constraints.NotNull;
-
-/**
- * The generic task message class to be serialised from or to a queue. This will contain the serialised worker-specific data inside.
- */
-public final class QueueTaskMessage
-{
-    public static final int TASK_MESSAGE_VERSION = 4;
```
dermot-hardy, line 29: Dropping support for the V4 format message is technically a breaking change so we should note it in the release notes.

PR #164, opened 2023-11-16
US857114: Move all AMQP communication to use TLS

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=857114


util-rabbitmq/src/main/java/com/hpe/caf/util/rabbitmq/RabbitUtil.java
```diff
@@ -93,13 +96,26 @@ public static Connection createRabbitConnection(final RabbitConfiguration rc) th
...
+            throws IOException, TimeoutException, URISyntaxException, NoSuchAlgorithmException, KeyManagementException {
         final ConnectionFactory factory = new ConnectionFactory();
         factory.setUsername(rc.getRabbitUser());
         factory.setPassword(rc.getRabbitPassword());
-        factory.setHost(rc.getRabbitHost());
-        factory.setPort(rc.getRabbitPort());
+
+        if (rc.getRabbitUrl() != null) {
+            factory.setUri(rc.getRabbitUrl());
+            if (!rc.getRabbitHost().equals("rabbitmq")) {
```
jonny-mccoubrey: @andyreidz have stopped using environment variables as we discussed and am checking that IF host and port aren't set to default then we output the warning.
However, this doesn't account for if host or port have been set to the default value. What do you think?

andyreidz: This could be confusing, perhaps we can change the logic in the default config js, so we don't default host and port when URL is set?
Then you can check host and port here safe in the knowledge that if they have values they're not the defaults and we need to warn.

andyreidz (changes requested):
Forgot to ask for release notes update

andyreidz (approved):

PR #175, opened 2024-01-23
I445035: Use dropwizard lifecycle for graceful shutdown

andyreidz:
Use dropwizard lifecycle for shutdown, worker will be allowed to continue to process pre-fetched messages.

### DIFF
--- a/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
+++ b/worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
@@ -89,7 +89,7 @@ public final class WorkerApplication extends Application<WorkerConfiguration>
     public void run(final WorkerConfiguration workerConfiguration, final Environment environment)
         throws QueueException, ModuleLoaderException, CipherException, ConfigurationException, DataStoreException, WorkerException
     {
-        LOG.debug("Worker initializing.");
+        LOG.info("Worker initializing.");
 
         ResponseCache.setDefault(new JobStatusResponseCache());
         BootstrapConfiguration bootstrap = new SystemBootstrapConfiguration();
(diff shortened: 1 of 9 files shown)


andyreidz:
@dermot-hardy I will create another ticket to cover setting a grace period for slow workers in the k8s deployments.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
```diff
@@ -108,13 +109,37 @@ public void run(final WorkerConfiguration workerConfiguration, final Environment
...
+                LOG.debug("Worker stop requested, allowing in-progress tasks to complete.");
                 workerQueue.shutdownIncoming();
+                while (!wtp.isIdle()) {
+                    try {
+                        //The grace period will expire and the process killed so no need for time limit here
+                        LOG.trace("Awaiting the Worker Thread Pool to become idle, {} tasks in the backlog.", 
+                                wtp.getBacklogSize());
+                        Thread.sleep(1000);
+                    } catch (final InterruptedException e) {
+                        throw new RuntimeException(e);
```
dermot-hardy, line 140: I know it's not going to make a lot of difference in this case but you should be resetting the interrupt flag before re-throwing.

```suggestion
                        Thread.currentThread().interrupt();
                        throw new RuntimeException(e);
```

worker-test/pom.xml
```diff
@@ -38,6 +38,11 @@
             <groupId>com.github.cafapi</groupId>
             <artifactId>caf-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.cafapi.logging</groupId>
+            <artifactId>caf-logging-logback</artifactId>
+            <scope>runtime</scope>
+        </dependency>
```
dermot-hardy: Move this down with the other runtime-scope dependencies.

dermot-hardy (approved):

michael-bryson (approved):

PR #186, opened 2024-05-03
US915147: Add support for liveness + readiness checks

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=915147

- The `WorkerFactory` interface contains a new `livenessCheck` method, which has a default implementation that returns 
`HealthResult.RESULT_HEALTHY`. A worker may optionally override this method to provide their own implementation of liveness.
- A new `/health-check?name=all&type=ALIVE` endpoint has been added on the default REST port (8080) to check if a worker is alive
- A new `/health-check?name=all&type=READY` endpoint has been added on the default REST port (8080) to check if a worker is ready
- See the [documentation](https://github.com/WorkerFramework/worker-framework/tree/develop/worker-core#health-checks-within-the-worker-framework) for more details.


release-notes-9.0.0.md
```diff
@@ -0,0 +1,20 @@
+!not-ready-for-release!
```
dermot-hardy: The 8.1.0 release notes file is still present.

worker-store-http/src/main/java/com/hpe/caf/worker/datastore/http/HttpDataStore.java
```diff
@@ -74,7 +74,13 @@ public void shutdown()
     }
 
     @Override
-    public HealthResult healthCheck()
+    public HealthResult checkAlive()
+    {
+        return new HealthResult(HealthStatus.HEALTHY);
+    }
```
dermot-hardy: As discussed we've decided to make this a default implementation at the `caf-common` level so this shouldn't be needed here.

dermot-hardy (changes requested):

worker-core/readme.md
```diff
@@ -580,7 +683,6 @@ the current input queue. Default is True.
       return resultQueue;
     }
 
-
```
dermot-hardy: Don't drop this blank line.  There are two blank lines between all the other methods.

worker-test/pom.xml
```diff
@@ -174,7 +174,7 @@
                             <alias>keystore</alias>
                             <name>${project.artifactId}-keystore:${project.version}</name>
                             <build>
-                                <from>${projectDockerRegistry}/cafapi/opensuse-jre11</from>
+                                <from>${dockerHubPublic}/cafapi/opensuse-jre11:3.0.0</from>
```
dermot-hardy: This doesn't seem right.  Is it a merge issue?

worker-test/pom.xml
```diff
@@ -194,7 +194,7 @@
                         </image>
                         <image>
                             <alias>webdav</alias>
-                            <name>${projectDockerRegistry}/cloudesire/webdav</name>
+                            <name>${dockerHubPublic}/cloudesire/webdav</name>
```
dermot-hardy: This doesn't seem right.  Is it a merge issue?

worker-test/pom.xml
```diff
@@ -254,14 +254,14 @@
                             <alias>worker-test</alias>
                             <name>${targetDockerRegistryPath}/worker-test:${project.version}</name>
                             <build>
-                                <from>${projectDockerRegistry}/cafapi/opensuse-jre17</from>
+                                <from>${dockerHubPublic}/cafapi/opensuse-jre17:1</from>
```
dermot-hardy: This doesn't seem right.  Is it a merge issue?

worker-test/pom.xml
```diff
@@ -340,6 +340,7 @@
                                 <ports>
                                     <port>${worker.testadminport2}:8081</port>
                                     <port>${worker.testdebugport2}:5005</port>
+                                    <port>8999:8080</port>
```
dermot-hardy: We shouldn't use fixed ports in the build as they will lead to conflicts on the build machine is the port is already in use.

worker-configs/src/main/java/com/hpe/caf/configs/HealthConfiguration.java
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * The initial delay to use when first scheduling the liveness check.
+     */
+    @Min(0)
+    private int livenessInitialDelaySeconds = 15;
+
+    /**
+     * The interval on which to perform a liveness check for while in a healthy state.
+     */
+    @Min(60)
```
dermot-hardy: Doesn't feel right for the minimum to be 60 here.  Same for a number of other configs also.

worker-core/src/main/java/com/hpe/caf/worker/core/WorkerApplication.java
```diff
@@ -107,8 +118,10 @@ public void run(final WorkerConfiguration workerConfiguration, final Environment
         WorkerThreadPool wtp = WorkerThreadPool.create(workerFactory);
         final int nThreads = workerFactory.getWorkerThreads();
         ManagedWorkerQueue workerQueue = queueProvider.getWorkerQueue(config, nThreads);
+
```
dermot-hardy:
```suggestion
```

dermot-hardy (changes requested):

worker-configs/src/main/java/com/hpe/caf/configs/HealthConfiguration.java
rorytorneymf: I didn't want to allow users to be able to set this to a very low value, as that would result in the checks getting run continuously, which I though could be a performance problem. I choose 60 as I think that is the lowest we currently use for `healthChecks.intervalSeconds` in Verity/deploy. What would be a better minimum value here, 10 seconds?

dermot-hardy: I'd take it all the way down to 1 or 2.  It is just a minimum &mdash; it's not like you're advising that configuration.

dermot-hardy (approved):

PR #189, opened 2024-05-16
I916311: Drop digest for dynamic images

kusumaghoshdastidar:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=916311


dermot-hardy (changes requested):
This won't last as a fix.  Clearly `2.4.59-bookworm` must not be a fixed tag.  We need to drop the digest.

dermot-hardy (approved):

PR #199, opened 2024-08-13
US948127: Investigate "Worker did not fail liveness check when out of memory"

jonny-mccoubrey:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=948127

### DIFF
--- a/worker-test/src/main/config/worker.sh
+++ b/worker-test/src/main/config/worker.sh
@@ -34,6 +34,7 @@ fi
 
 cd /maven
 exec java $CAF_WORKER_JAVA_OPTS \
+    -Dpolyglot.engine.WarnInterpreterOnly=false \
     -cp '*' \
     com.hpe.caf.worker.core.WorkerApplication \
     server \
--- a/worker-test/src/main/config/worker.sh
+++ b/worker-test/src/main/config/worker.sh
@@ -26,7 +26,6 @@ export HOSTNAME=worker-test
 
 cd /maven
 exec java $CAF_WORKER_JAVA_OPTS \
-    -XX:+CrashOnOutOfMemoryError \
     -cp '*' \
     com.hpe.caf.worker.core.WorkerApplication \
     server \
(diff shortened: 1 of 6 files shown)


worker-test/src/main/config/worker.sh
```diff
@@ -22,9 +22,18 @@ export CAF_CONFIG_ENABLE_SUBSTITUTOR=false
 
 export CAF_WORKER_DISABLE_ZERO_PROGRESS_REPORTING=true
 
+if [ "$CRASH_DUMP_ON_OUT_OF_MEMORY_ERROR" == "true" ]
```
jonny-mccoubrey, line 25: @dermot-hardy what do you think about this in the `worker.sh` script?

worker-test/src/main/config/worker.sh
```diff
@@ -22,9 +22,18 @@ export CAF_CONFIG_ENABLE_SUBSTITUTOR=false
...
+fi
+
+if [ "$HEAP_DUMP_ON_OUT_OF_MEMORY_ERROR" == "true" ]
+then
+  CAF_WORKER_JAVA_OPTS="${CAF_WORKER_JAVA_OPTS} -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=${DUMP_PATH}heap_dump_${HOSTNAME}.hprof"
+fi
+
 cd /maven
 exec java $CAF_WORKER_JAVA_OPTS \
-    -Dpolyglot.engine.WarnInterpreterOnly=false \
```
dermot-hardy, line 27: Why are you removing this line?

jonny-mccoubrey: That's a mistake, sorry

worker-test/src/main/config/worker.sh
dermot-hardy: It seems reasonable.

PR #203, opened 2024-10-08
US969005: Support for reading secrets

rorytorneymf:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=969005

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -213,42 +213,42 @@
             <dependency>
                 <groupId>com.github.cafapi</groupId>
                 <artifactId>caf-api</artifactId>
-                <version>3.2.0-US969005-2-SNAPSHOT</version>
+                <version>3.2.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.cipher</groupId>
                 <artifactId>cipher-null</artifactId>
-                <version>3.2.0-US969005-2-SNAPSHOT</version>
+                <version>3.2.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.codec</groupId>
                 <artifactId>codec-json</artifactId>
-                <version>3.2.0-US969005-2-SNAPSHOT</version>
+                <version>3.2.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.config</groupId>
                 <artifactId>config-caf</artifactId>
-                <version>3.2.0-US969005-2-SNAPSHOT</version>
+                <version>3.2.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.config</groupId>
                 <artifactId>config-decoder</artifactId>
-                <version>3.2.0-US969005-2-SNAPSHOT</version>
+                <version>3.2.0-SNAPSHOT</version>
             </dependency>
             <dependency>
                 <groupId>com.github.cafapi.config</groupId>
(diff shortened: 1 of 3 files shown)


worker-default-configs/config/cfg~caf~worker~RabbitConfiguration.js
```diff
@@ -21,5 +21,5 @@
     rabbitHost: getenv("CAF_RABBITMQ_HOST") || "rabbitmq",
     rabbitPort: getenv("CAF_RABBITMQ_PORT") || 5672,
     rabbitUser: getenv("CAF_RABBITMQ_USERNAME") || "guest",
-    rabbitPassword: getenv("CAF_RABBITMQ_PASSWORD") || "guest"
+    rabbitPassword: getenvorprop("CAF_RABBITMQ_PASSWORD", "CAF.") || "guest"
```
andyreidz: Could we just use getprop, since all envs will be convert to properties?

rorytorneymf: I think that would be a breaking change, as currently we support just providing the `CAF_RABBITMQ_PASSWORD` env var (i.e. just as it is, not as a converted file-based secret) - lots of the integration tests do this.

If were to change this to getprop only, then it forces all workers to use this new property-based mechanism, and would break any worker that just passed `CAF_RABBITMQ_PASSWORD=something`.

Using `getenvorprop` avoids the breaking change.

rorytorneymf: Maybe a clearer way to implement this might be:

```js
rabbitPassword: getenv("CAF_RABBITMQ_PASSWORD") || getprop("CAF.CAF_RABBITMQ_PASSWORD") || "guest"
```

dermot-hardy: If you create a new function like `getSecret()` then you can encapsulate the logic inside it:

```suggestion
    rabbitPassword: getSecret("CAF_RABBITMQ_PASSWORD") || "guest"
```

So that function could then decide internally whether to get the secret directly from the environment variable, or fallback to appending `_FILE` and reading it from disk, or fallback to prepending `CAF.` and reading it from system property, or in future perhaps fallback to reading it directly from Vault, ...

rorytorneymf: Yeah that sounds good to me, I'll update the PRs and see what that looks like

pom.xml
```diff
@@ -213,42 +213,42 @@
             <dependency>
                 <groupId>com.github.cafapi</groupId>
                 <artifactId>caf-api</artifactId>
-                <version>3.1.0-596</version>
+                <version>3.2.0-US969005-2-SNAPSHOT</version>
```
rorytorneymf: Update versions before merging.

worker-default-configs/config/cfg~caf~worker~RabbitConfiguration.js
rorytorneymf: Looks to work well thanks, I'll assign out the initial MRs for proper review.

rorytorneymf: Sorry, I didn't see your comment on the ticket regarding these approaches, I'll reply on the ticket now

andyreidz (approved):

andyreidz (approved):

dermot-hardy (approved):

PR #214, opened 2025-04-16
US1009117: Remove reliance on large messages being supported

DavidGerald-Milligan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1009117


worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
```diff
@@ -151,12 +151,11 @@ public CoreTaskCallback(final Codec codec, final WorkerStats stats, final Worker
          * Use the factory to get a new worker to handle the task, wrap this in a handler and hand it off to the thread pool.
          */
         @Override
-        public void registerNewTask(final TaskInformation taskInformation, final byte[] taskMessage, Map<String, Object> headers)
+        public void registerNewTask(final TaskInformation taskInformation, final TaskMessage taskMessage, Map<String, Object> headers)
             throws InvalidTaskException, TaskRejectedException
         {
             Objects.requireNonNull(taskInformation);
             stats.incrementTasksReceived();
-            stats.getInputSizes().update(taskMessage.length);
```
DavidGerald-Milligan, line 159: Discussed and identified as unused, removed to save deserializing the taskmessage again.

worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
```diff
@@ -219,9 +216,7 @@ private void registerNewTaskImpl(final TaskInformation taskInformation, final by
                             taskInformation.getInboundMessageId());
                         executor.discardTask(tm, taskInformation);
                 }
-            } catch (CodecException e) {
-                throw new InvalidTaskException("Queue data did not deserialise to a TaskMessage", e);
-            } catch (InvalidJobTaskIdException ijte) {
+            } catch (final InvalidJobTaskIdException ijte) {
                 throw new InvalidTaskException("TaskMessage contains an invalid job task identifier", ijte);
             }
         }
```
DavidGerald-Milligan, line 223: Desialization now done earlier in WorkerQueueConsumerImpl.java

util-rabbitmq/src/main/java/com/github/workerframework/util/rabbitmq/RabbitHeaders.java
```diff
@@ -23,4 +23,5 @@ public class RabbitHeaders
     public static final String RABBIT_HEADER_CAF_WORKER_REJECTED = "x-caf-worker-rejected";
     public static final String RABBIT_HEADER_CAF_WORKER_RETRY = "x-caf-worker-retry";
     public static final String RABBIT_HEADER_CAF_DELIVERY_COUNT = "x-delivery-count";
+    public static final String RABBIT_HEADER_CAF_DEHYDRATION_ID = "x-dehydration-id";
```
andyreidz:
```suggestion
    public static final String RABBIT_HEADER_CAF_DEHYDRATION_ID = "x-caf-dehydration-id";
```

worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
```diff
@@ -151,12 +151,11 @@ public CoreTaskCallback(final Codec codec, final WorkerStats stats, final Worker
          * Use the factory to get a new worker to handle the task, wrap this in a handler and hand it off to the thread pool.
          */
         @Override
-        public void registerNewTask(final TaskInformation taskInformation, final byte[] taskMessage, Map<String, Object> headers)
+        public void registerNewTask(final TaskInformation taskInformation, final TaskMessage taskMessage, Map<String, Object> headers)
```
andyreidz, line 152: import com.github.cafapi.common.api.DecodeMethod; is unused

worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
andyreidz: Remove from WorkerApplication.

        metrics.register(MetricRegistry.name("core.inputSizes"), core.getStats().getInputSizes());

and from WorkerStats

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitTaskInformation.java
```diff
@@ -30,19 +31,47 @@ public class RabbitTaskInformation implements TaskInformation {
...
 
     public RabbitTaskInformation(final String inboundMessageId) {
         this(inboundMessageId, false);
     }
     
     public RabbitTaskInformation(final String inboundMessageId, final boolean isPoison) {
+        this(inboundMessageId, isPoison, Optional.empty());
+    }
+
+    public RabbitTaskInformation(final String inboundMessageId, final boolean isPoison, final Optional<String> dehydratedMessageId) {
```
andyreidz:
```suggestion
    public RabbitTaskInformation(final String inboundMessageId, final boolean isPoison, final Optional<String> dehydratedTaskMessageStorageRef) {
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
```diff
@@ -66,18 +86,47 @@ public void handlePublish(byte[] data, String routingKey, RabbitTaskInformation
...
+    }
+
+    private byte[] getOutboundByteArray(
+        final byte[] taskMessage,
+        final String routingKey,
+        final Map<String, Object> headers
+    ) throws QueueException {
+        try {
+            if (shouldStoreTaskMessage(taskMessage.length)) {
+                final TaskMessage outboundTaskMessage = codec.deserialise(taskMessage, TaskMessage.class);
```
andyreidz: Change handlePublish to accept TaskMessage instead of byte[] so we avoid deserialization here.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -180,14 +227,17 @@ private void processReject(long id, boolean requeue)
...
 
         final RabbitTaskInformation taskInformation = 
                 new RabbitTaskInformation(String.valueOf(delivery.getEnvelope().getDeliveryTag()));
         LOG.debug("Received redelivered message with id {}, retry count {}, retry limit {}, republishing to retry queue",
                 delivery.getEnvelope().getDeliveryTag(), retryLimit, retries + 1);
         final Map<String, Object> headers = new HashMap<>();
         headers.put(RabbitHeaders.RABBIT_HEADER_CAF_WORKER_RETRY, String.valueOf(retries + 1));
+        if (dehydratedMessageId.isPresent()) {
+            headers.put(RABBIT_HEADER_CAF_DEHYDRATION_ID, dehydratedMessageId.get());
+        }
```
andyreidz: For a poison message, using classic queues, won't the dehydratedMessageId be deleted when the delivered message is ack'd?

Also the delivery.getMessageData() will return an empty byte array for a dehydrated message?

andyreidz (changes requested):
Main thing is poison message handling which will need testing for both quorum and classic queue modes, it's likely that classic is problem area because it publishes a new message which is a copy of the original with a modified header.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
DavidGerald-Milligan: @andyreidz I see why you say this but this creates an issue in the consumer if we cannot deserialize the incoming byte array to a TaskMessage and have to republish.  Is it a valid option to reject the delivery at this point as we do in handlePublish if it cannot be deserialized? https://github.com/WorkerFramework/worker-framework/blob/US1009117/worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java#L135

DavidGerald-Milligan: actually, we have to either deserialize or serialize here anyway, both storing and publishing require one or other of the actions.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: _If we call republishClassicRedelivery, the incoming dehydratedMessageId is not set in the TaskInformation so it will not be deleted.   Yes the messageData will be empty if its a dehydrated message, but it will be re-hydrated on re-delivery._

**Update**: I didn't consider the message size on republish would mean the header is not set,  we need to read the stored message here pass that on for republishing and then delete the original message here, or find a way to pass the existing reference to the publisher.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -100,20 +118,72 @@ public void processDelivery(Delivery delivery)
...
-            LOG.debug("Registering new message {}", taskInformation.getInboundMessageId());
-            callback.registerNewTask(taskInformation, delivery.getMessageData(), delivery.getHeaders());
+            final TaskMessage taskMessage = deserializeTaskMessage(inboundMessageId, inboundByteArray, taskMessageStorageRefOpt);
+            final var trackingInfo = taskMessage.getTracking();
+            final var trackingJobTaskId = trackingInfo != null ? trackingInfo.getJobTaskId() : "untracked";
+            final RabbitTaskInformation taskInformation = new RabbitTaskInformation(
+                String.valueOf(inboundMessageId),
+                isPoison,
+                taskMessageStorageRefOpt,
+                Optional.of(String.format("%s/%s", routingKey, trackingJobTaskId))
```
andyreidz: Just store the trackingJobTaskId here.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
```diff
@@ -66,18 +82,52 @@ public void handlePublish(byte[] data, String routingKey, RabbitTaskInformation
...
+            final Optional<String> inboundTaskMessageStorageRef = Optional.ofNullable(
+                publishHeaders.get(RABBIT_HEADER_CAF_MINIMIZATION_ID)
+            ).map(Object::toString);
+
+            if (inboundTaskMessageStorageRef.isPresent() && taskInformation.getMinimizedTaskMessageStorageRef().isPresent()) {
+                // We have successfully rehydrated this message and the minimized message id is redundant.
+                // The stored message will be deleted in the confirm listener
+                publishHeaders.remove(RABBIT_HEADER_CAF_MINIMIZATION_ID);
+            }
+            final var outboundByteArray = getOutboundByteArray(data, taskInformation.getTaskMessagePartialRef(), publishHeaders);
```
andyreidz: taskInformation.getTaskMessagePartialRef is the queue the message was delivered to.

When publishing, the partial refererence should be constructed from the routingKey passed to handlePublish so we store the TaskMessage in a location associated with the publish queue, not the delivery queue.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitTaskInformation.java
```diff
@@ -30,19 +31,56 @@ public class RabbitTaskInformation implements TaskInformation {
     private final AtomicInteger acknowledgementCount;
     private static final Logger LOG = LoggerFactory.getLogger(RabbitTaskInformation.class);
     private final boolean isPoison;
+    private final Optional<String> minimizedTaskMessageStorageRef;
+    private final Optional<String> taskMessagePartialRef;
```
andyreidz: Sorry, I see now this was a bad idea.

Look in the WorkerPublisherImpl.handlePublish, there we're using this to store the TaskMessage bytes in the Datastore when we should be storing to the routingKey where we intend to send the message to.

Change taskMessagePartialRef to trackingInfoJobTaskId

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
```diff
@@ -66,18 +82,52 @@ public void handlePublish(byte[] data, String routingKey, RabbitTaskInformation
     {
         try {
             LOG.debug("Publishing message to {} with ack id {}", routingKey, taskInformation.getInboundMessageId());
+            final var publishHeaders = new HashMap<>(headers);
+            // Remove any previous minimization id
+            final Optional<String> inboundTaskMessageStorageRef = Optional.ofNullable(
```
andyreidz: Are there any downsides to simplifying this to always remove the RABBIT_HEADER_CAF_MINIMIZATION_ID header from publishHeaders?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
```diff
@@ -66,18 +82,52 @@ public void handlePublish(byte[] data, String routingKey, RabbitTaskInformation
...
+        try {
+            if (taskMessagePartialRef.isPresent() && shouldStoreTaskMessage(taskMessage.length)) {                
+                final var taskMessageStorageRef = dataStore.store(taskMessage, taskMessagePartialRef.get());
+                headers.put(RABBIT_HEADER_CAF_MINIMIZATION_ID, taskMessageStorageRef);
+                //  if the header is set, the consumer will ignore the incoming byte[] and use the minimized message.
+                return new byte[0];
+            }
+            return taskMessage;
+        } catch (final Exception e) {
+            throw new QueueException("Error minimizing task message", e);
```
andyreidz: Why are we catching Exception and mapping to QueueException?

Only DataStoreException can be thrown from that block, so it seems like we should let that propagate out to handlePublish and handle it there.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -177,17 +247,20 @@ private void processReject(long id, boolean requeue)
...
      * Republish the delivery to the retry queue with the retry count stamped in the headers.
-     *
-     * @param delivery the redelivered message
+     * 
+     * @param delivery The redelivered message
+     * @param retries
+     * @param taskMessageStorageRefOpt
      */
-    private void republishClassicRedelivery(final Delivery delivery, final int retries) {
+    private void republishClassicRedelivery(final Delivery delivery, final int retries, final Optional<String> taskMessageStorageRefOpt) {
```
andyreidz: We're reusing the storage reference from the delivered message when publishing to the configured retry queue. 

How are we preventing the WorkerConfirmListener.handleAck from deleting this storage reference when the message is published to the retry queue?

Additionally, the retry queue is usually configured to be the same as the input queue but it's possible that it could be a different queue entirely. 

If that is the case the message sent to that retry queue would have a header with a storage reference containing the original queue's name and not the name of the retry queue.

It's better to avoid reusing the storage reference from the delivered message just create a new copy using retryRoutingKey in the partial reference, then we don't have to worry about the confirm listener.

There's a comment in the ticket saying that we ask operators to purge reject queues. With this new functionality that process will also involve them deleting the location in storage associated with that queue.  That's why it's important that the storage reference in the header always aligns with the queue the message is sitting on.

We also need to think about how this works will message prioritization as the queue names with be staging queues, and it's message prioritization worker that moves them to the intended queue.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -100,20 +118,72 @@ public void processDelivery(Delivery delivery)
...
+            if (taskMessageStorageRefOpt.isPresent()) {
+                final ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+                try (final var inputStream = dataStore.retrieve(taskMessageStorageRefOpt.get())) {
+                    final byte[] buffer = new byte[1024];
+                    int length;
+                    while ((length = inputStream.read(buffer)) != -1) {
+                        outputStream.write(buffer, 0, length);
+                    }
+                }
+                return codec.deserialise(outputStream.toByteArray(), TaskMessage.class, DecodeMethod.LENIENT);
```
andyreidz:
```suggestion
               return codec.deserialise(deliveryMessageData, TaskMessage.class, DecodeMethod.LENIENT);
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -100,20 +118,72 @@ public void processDelivery(Delivery delivery)
...
+            }
+           return codec.deserialise(deliveryMessageData, TaskMessage.class, DecodeMethod.LENIENT);
+        } catch (final IOException | CodecException | DataStoreException e) {
```
andyreidz: If you apply the above suggestion then IOException can be removed.

```suggestion
        } catch (final CodecException | DataStoreException e) {
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerConfirmListener.java
```diff
@@ -123,4 +131,13 @@ private void handle(long sequenceNo, boolean multiple, Function<RabbitTaskInform
...
+
+    private void deleteMinimizedMessage(final String taskMessageStorageRef)
+    {
+        try {
+            dataStore.delete(taskMessageStorageRef);
+        } catch (final DataStoreException e) {
+            LOG.error("Failed to delete a minimized message:{} from the datastore", taskMessageStorageRef, e);
+        }
+    }
+}
```
andyreidz: newline

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -100,20 +118,72 @@ public void processDelivery(Delivery delivery)
...
+                String.valueOf(inboundMessageId),
+                isPoison,
+                taskMessageStorageRefOpt,
+                Optional.of(String.format("%s/%s", routingKey, trackingJobTaskId))
+            );                      
+            LOG.debug("Registering new message {}", inboundMessageId);
+            callback.registerNewTask(taskInformation, taskMessage, delivery.getHeaders());
         } catch (InvalidTaskException e) {
-            LOG.error("Cannot register new message, rejecting {}", taskInformation.getInboundMessageId(), e);
+            final RabbitTaskInformation taskInformation = new RabbitTaskInformation(
```
andyreidz: I made comments on republishClassicRedelivery about the dangers of reusing the storage reference when republishing a new message. If we stop doing that then we can these catch blocks can be simplified to the original approach with the exception of the required code to create a copy of the data in the datastore.

We should handle a problem retrieving from the datastore by adding a entirely new header to record the original storage reference and details of the exception.

andyreidz (changes requested):
Let's talk when you've have a chance to go through my comments.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
andyreidz: It might be better to capture the byte array into a variable that you can reuse it in the catch blocks when sending the WorkerPublishQueueEvents

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitTaskInformation.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
DavidGerald-Milligan: always removing this now

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: I think we are done here

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerConfirmListener.java
DavidGerald-Milligan: done

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerPublisherImpl.java
```diff
@@ -66,18 +82,40 @@ public void handlePublish(byte[] data, String routingKey, RabbitTaskInformation
...
+    ) throws DataStoreException {
+        headers.remove(RABBIT_HEADER_CAF_MINIMIZATION_ID);
+        if (trackingJobTaskId.isPresent() && shouldStoreTaskMessage(taskMessage.length)) {
+            final var taskMessageStorageRef = dataStore.store(taskMessage, String.format("%s/%s", routingKey, trackingJobTaskId.get()));
+            headers.put(RABBIT_HEADER_CAF_MINIMIZATION_ID, taskMessageStorageRef);
+            //  if the header is set, the consumer will ignore the incoming byte[] and use the minimized message.
+            return new byte[0];
+        }
+        return taskMessage;
+    }
```
DavidGerald-Milligan: note to self: republishClassicRedelivery does not supply trackingJobTaskId

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -29,67 +36,133 @@
...
 
         metrics.incrementReceived();
+
+        // If the message is minimized, we need to retrieve it from the data store.
+        final byte[] taskMessageData;
+        if (taskMessageStorageRefOpt.isPresent()) {
+            try {
+                taskMessageData = retrieveFromDatastore(taskMessageStorageRefOpt.get());
+            } catch (final IOException | DataStoreException e) {
+                // The message was minimized, but we could not retrieve it from the data store.
```
andyreidz: Extract this catch block and the one starting on ln139 into a method, they seem to be the same.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -100,21 +173,62 @@ public void processDelivery(Delivery delivery)
...
+        } catch (final TaskRejectedException e) {
+            LOG.warn("Message {} rejected as a task at this time, returning to queue", inboundMessageId, e);
             taskInformation.incrementResponseCount(true);
-            publisherEventQueue.add(new WorkerPublishQueueEvent(delivery.getMessageData(), delivery.getEnvelope().getRoutingKey(),
-                    taskInformation));
+            publisherEventQueue.add(new WorkerPublishQueueEvent(taskMessageByteArray, routingKey, taskInformation, deliveryHeaders));
+        }
+    }
+
+    private byte[] retrieveFromDatastore(final String taskMessageStorageRef) throws DataStoreException, IOException {
```
andyreidz: I don't think we need this buffering.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -29,67 +36,133 @@
...
+        final Optional<String> taskMessageStorageRefOpt = Optional.ofNullable(
+            deliveryHeaders.get(RABBIT_HEADER_CAF_MINIMIZATION_ID)
+        ).map(Object::toString);
 
         metrics.incrementReceived();
+
+        // If the message is minimized, we need to retrieve it from the data store.
+        final byte[] taskMessageData;
+        if (taskMessageStorageRefOpt.isPresent()) {
+            try {
```
andyreidz:
```suggestion
            try (final var inputStream = dataStore.retrieve(taskMessageStorageRefOpt.get())) {
                taskMessageData = inputStream.readAllBytes();
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -29,67 +36,133 @@
...
+                taskMessageData = retrieveFromDatastore(taskMessageStorageRefOpt.get());
```
andyreidz:
```suggestion
```

andyreidz (approved):
Let's discuss at standup when to merge this.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -44,76 +56,189 @@
...
+        final Optional<String> taskMessageStorageRefOpt,
+        final long inboundMessageId
+    ) {
+        if (taskMessageStorageRefOpt.isPresent()) {
+            final String taskMessageStorageRef = taskMessageStorageRefOpt.get();  
+            offloadedPayloads.put(inboundMessageId, taskMessageStorageRef);
+            try (final var inputStream = dataStore.retrieve(taskMessageStorageRef)) {
+                return inputStream.readAllBytes();
+            } catch (final IOException | DataStoreException e) {
+                final RabbitTaskInformation taskInformation = new RabbitTaskInformation(String.valueOf(inboundMessageId), true);
```
dermot-hardy: This message didn't crash the worker so why are we declaring it to be a poison message?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -44,76 +56,189 @@
...
+    ) {
+        try {
+            return codec.deserialise(taskMessageData, TaskMessage.class, DecodeMethod.LENIENT);
+        } catch (final CodecException e) {
+            final RabbitTaskInformation errorTaskInformation = new RabbitTaskInformation(String.valueOf(inboundMessageId), true);
+            LOG.error("Cannot register new message, rejecting {}", inboundMessageId, e);
+            errorTaskInformation.incrementResponseCount(true);
+            final var publishHeaders = new HashMap<String, Object>();
+            publishHeaders.put(RabbitHeaders.RABBIT_HEADER_CAF_WORKER_REJECTED, REJECTED_REASON_TASKMESSAGE);
+            publishHeaders.put(RabbitHeaders.RABBIT_HEADER_CAF_PAYLOAD_OFFLOADING_STORAGE_REF, taskMessageStorageRefOpt.orElse(null));
```
dermot-hardy: Is setting this header to null safe?  Shouldn't it just not be set?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -44,76 +56,189 @@
...
+    private void processDelivery(
+        final long inboundMessageId,
+        final String routingKey,
+        final Map<String, Object> deliveryHeaders,
+        final TaskMessage taskMessage,
+        final byte[] taskMessageByteArray,
+        final boolean isPoison
+    ) {
+        final TrackingInfo trackingInfo = taskMessage.getTracking();
+        final String trackingJobTaskId = trackingInfo != null ? trackingInfo.getJobTaskId() : "untracked";
```
dermot-hardy, line 333: What does a normal job task id look like?  Would there be any confusion if someone created a job with "untracked" as the job id?  Presumably not because the partition is in there somewhere?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -44,76 +56,189 @@
...
-            LOG.warn("Message {} rejected as a task at this time, returning to queue", taskInformation.getInboundMessageId(), e);
+            final var publishHeaders = new HashMap<String, Object>();
+            publishHeaders.put(RabbitHeaders.RABBIT_HEADER_CAF_WORKER_REJECTED, REJECTED_REASON_TASKMESSAGE);
+            publisherEventQueue.add(new WorkerPublishQueueEvent(taskMessageByteArray, retryRoutingKey, taskInformation, publishHeaders));
+        } catch (final TaskRejectedException e) {
+            LOG.warn("Message {} rejected as a task at this time, returning to queue", inboundMessageId, e);
             taskInformation.incrementResponseCount(true);
-            publisherEventQueue.add(new WorkerPublishQueueEvent(delivery.getMessageData(), delivery.getEnvelope().getRoutingKey(),
-                    taskInformation));
+            publisherEventQueue.add(new WorkerPublishQueueEvent(taskMessageByteArray, routingKey, taskInformation, deliveryHeaders));
```
dermot-hardy, line 349: I notice the `deliveryHeaders` are being added here in this rejection circumstance?  Is that intentional?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -131,6 +256,17 @@ public void processAck(long tag)
             LOG.warn("Couldn't ack message {}, will retry", tag, e);
             metrics.incremementErrors();
             consumerEventQueue.add(new ConsumerAckEvent(tag));
+            return;
+        }
+
+        final String datastorePayloadReference = offloadedPayloads.get(tag);
```
dermot-hardy: I see we're deleting the payload from the datastore here, but we don't seem to ever be removing the entry from the `offloadedPayloads` map.  This seems like a memory leak.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: Yes this would include the offloading header which, if present, would still be required when processing the rejected message.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: https://github.com/WorkerFramework/worker-framework/pull/214/commits/0fd03112e24dec1914a26eac871fa38eed41fcd3

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: https://github.com/WorkerFramework/worker-framework/pull/214/commits/8e66c073e9d6b49adce98113a2976740cccf6c39

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: The intent was to easily spot orphaned offloaded payloads in the datastore.  Would the confusion lie in job monitoring?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: I think this was to send it directly to the reject queue since the expected message was not in the data store.  @andyreidz  ?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
dermot-hardy: What does the actual directory structure look like?

DavidGerald-Milligan: Where this was tested on larry:
/etc/store/dataprocessing-update-entities-in/tenant-dave-001\:eca_g_503e01c3-e429-4b01-828c-90e04a8c058b.3.1.1/3e09ed9a-1c7d-4f71-8436-949f7566321d

where 3e09ed9a-1c7d-4f71-8436-949f7566321d is the offloaded payload

dermot-hardy: I took a look on Larry to see what this looks like:
image

It seems worrying that the queues are at the same directory level as the `batches` directory and the `device...` directories.  Shouldn't there be another directory level coming before this.

dermot-hardy: Also looking inside the `dataprocessing-update-entities-in` folder I see the following &mdash; a directory for each task:

    ├── tenant-dave-001:eca_g_2b783514-9c3f-4e4d-a866-eb247ab590df.2.1.1
    ├── tenant-dave-001:eca_g_44fabc03-b023-49e2-a852-fd5d4ec4ed45.5.1.1
    ├── tenant-dave-001:eca_g_503e01c3-e429-4b01-828c-90e04a8c058b.3.1.1
    ├── tenant-dave-001:eca_g_61ff8eff-163a-4ca0-a82a-4de956e0b5c1.2.1.1
    └── tenant-dave-001:eca_g_682de301-7ba4-4dc6-8dc1-74d65c3c3f1f.1.1.1

Ultimately this will fail, because there will be too many directories under this one node.

It needs to be split out, ideally by tenant and then job id, so a directory structure more like this:

    └───tenant-dave-001
        ├───eca_g_2b783514-9c3f-4e4d-a866-eb247ab590df
        │   └───2.1.1
        ├───eca_g_44fabc03-b023-49e2-a852-fd5d4ec4ed45
        │   └───5.1.1
        ├───eca_g_503e01c3-e429-4b01-828c-90e04a8c058b
        │   └───3.1.1
        ├───eca_g_61ff8eff-163a-4ca0-a82a-4de956e0b5c1
        │   └───2.1.1
        └───eca_g_682de301-7ba4-4dc6-8dc1-74d65c3c3f1f
            └───1.1.1

andyreidz: Let's introduce an additional configuration item to control the root directory to address Dermot's point about the queue directories being at the same level as where the staging service batches are.

I suggest using "queues" as the default value.

andyreidz: There's something wrong with this.

image

It looks like the rest of the queue name after » is missing.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
andyreidz: Yes if we fail to retrieve from the datastore we treat it as poison. 

In com.github.cafdataprocessing.workers.document.fieldvalues.ReferenceFieldValue we throw IOException if there's a problem retrieving from DataStore.

In Entity Extract com.hpe.caf.worker.entityextract.core.CachedInputData.initialize we catch IOException and throw RuntimeException recording it as a Failure on the document.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: The dataprocessing-classification-in>> folder is not created by the offloading which was not enabled on worker-classification

DavidGerald-Milligan: The queue name is listed in RabbitMQ as
image

and the queue directory is created as
image

andyreidz: Ahh, of course.

DavidGerald-Milligan: The offloading directory has been updated as suggested

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
andyreidz: Updated to check for a ReferenceNotFoundException which is the only non-transient type of error. Only this case is treated as a poison message.

In other cases disconnectIncoming is triggered.

Thoughts @dermot-hardy ?

dermot-hardy: That sounds reasonable.  Later when we have enough of the workers updated we should test it in some real workflow scenarios to make sure it behaves as expected.

andyreidz:
@DavidGerald-Milligan Please review the latest commit, for testing we want to verify that transient and fatal datastore exceptions are handled correctly. Any associated job should be marked as encountering an error.

You can test fatal errors by deleting one of the offloaded payloads from the DataStore while running through a workflow. I haven't thought of an easy way to trigger a transient error, such as the DataStore being temporarily unavailable, let me know if you think of something.

andyreidz (approved):
@dermot-hardy would also like your review.

worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
```diff
@@ -151,12 +149,12 @@ public CoreTaskCallback(final Codec codec, final WorkerStats stats, final Worker
...
             throws InvalidTaskException, TaskRejectedException
         {
             Objects.requireNonNull(taskInformation);
             stats.incrementTasksReceived();
-            stats.getInputSizes().update(taskMessage.length);
+            stats.getInputSizes().update(taskMessage.getTaskData().length);
```
dermot-hardy: Are we sure that `.getTaskData()` never returns `null` at this point?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -34,86 +42,234 @@
...
-            LOG.debug("Registering new message {}", taskInformation.getInboundMessageId());
-            callback.registerNewTask(taskInformation, delivery.getMessageData(), delivery.getHeaders());
-        } catch (InvalidTaskException e) {
-            LOG.error("Cannot register new message, rejecting {}", taskInformation.getInboundMessageId(), e);
+            LOG.debug("Registering new message {}", inboundMessageId);
+            callback.registerNewTask(taskInformation, taskMessage, deliveryHeaders);
+        } catch (final InvalidTaskException e) {
+            LOG.error("Cannot register new message, rejecting {}", inboundMessageId, e);
             taskInformation.incrementResponseCount(true);
-            publisherEventQueue.add(new WorkerPublishQueueEvent(delivery.getMessageData(), retryRoutingKey, taskInformation,
```
andyreidz, line 110: There is an existing fault here. 

When an InvalidTaskException is caught we're sending a new message to the retryRoutingKey, which is typically the input queue and we'd get stuck in an infinite loop.

Looking at how InvalidTaskException is handled elsewhere I'm now passing this queue name through to the WorkerQueueConsumerImpl

https://github.com/search?q=InvalidTaskQueue+caf&type=code&p=1
https://gitlab.otxlab.net/search?group_id=86711&scope=blobs&search=InvalidTaskQueue

**Document Workers have CAF_WORKER_FAILURE_QUEUE which is mapped to InvalidTaskQueue**

https://github.com/CAFDataProcessing/worker-document/blob/200253ea9fad190866e98b22119c19dd8eb0827a/worker-document/src/main/java/com/github/cafdataprocessing/workers/document/config/DocumentWorkerConfiguration.java#L76

https://github.com/CAFDataProcessing/worker-document/blob/200253ea9fad190866e98b22119c19dd8eb0827a/worker-document/src/main/java/com/github/cafdataprocessing/workers/document/DocumentWorkerAdapter.java#L108

**Production service is mapping OutputQueue to InvalidTaskQueue**

https://gitlab.otxlab.net/cybersecurity-enterprise/cddri/app/production/-/blob/master/production-service/src/main/java/com/hpe/verity/worker/production/ProductionServiceFactory.java

**Worker markup is using outputQueue**

https://github.com/CAFDataProcessing/worker-markup/blob/f5b2e1e6d4fc5f82b805ccdc670da8d30e3dc01d/worker-markup/src/main/java/com/github/cafdataprocessing/workers/markup/MarkupWorkerFactory.java#L84

**Worker batch is using outputQueue**

https://github.com/JobService/worker-batch/blob/f4da1bc2a089a76d7eefc1f13447348c99899fb1/worker-batch/src/main/java/com/github/jobservice/workers/batch/BatchWorkerFactory.java#L74

**Worker task stowing is using failureQueue**

https://github.com/JobService/worker-taskstowing/blob/7b80911ed0cd6fd8bc09bc741e0761f29a4f947d/worker-taskstowing/src/main/java/com/github/jobservice/workers/taskstowing/factory/TaskStowingWorkerFactory.java#L77

**Job service is using outputqueue**

https://github.com/JobService/job-service/blob/0d8f2942e7feb6de655e4d1601b18323155ccb7f/worker-jobtracking/src/main/java/com/github/jobservice/workers/jobtracking/JobTrackingWorkerFactory.java#L243

**Example of existing usage**

```
StreamingWorkerWrapper.run()

catch (InvalidTaskException e) {
            workerTask.setResponse(e);
```


`WorkerTaskImpl.setResponse(final InvalidTaskException invalidTaskException)`

```
        final TaskMessage invalidResponse = new TaskMessage(
            MoreObjects.firstNonNull(taskMessage.getTaskId(), ""),
            taskClassifier,
            taskMessage.getTaskApiVersion(),
            taskData,
            TaskStatus.INVALID_TASK,
            context,
**            workerFactory.getInvalidTaskQueue(), **
```

worker-api/src/main/java/com/github/workerframework/api/WorkerQueueProvider.java
```diff
@@ -27,9 +28,11 @@ public interface WorkerQueueProvider
      *
      * @param configurationSource used for configuring the WorkerQueue
      * @param maxTasks the maximum number of tasks the worker can perform at once
+     * @param dataStore the managed data store that the worker will use to store data that exceeds a threshold.
+     * @param codec the codec used for serialization deserialization of data.
      * @return a new WorkerQueue instance
      * @throws QueueException if a WorkerQueue could not be created
      */
-    ManagedWorkerQueue getWorkerQueue(ConfigurationSource configurationSource, int maxTasks)
+    ManagedWorkerQueue getWorkerQueue(ConfigurationSource configurationSource, int maxTasks, String invalidQueue, ManagedDataStore dataStore, Codec codec)
```
dermot-hardy: Missing Javadoc param for `invalidQueue` and line greater than 138 characters long:

image

worker-core/src/main/java/com/github/workerframework/core/WorkerCore.java
```diff
@@ -502,17 +487,16 @@ public void complete(final TaskInformation taskInformation, final String queue,
                 } else {
                     // **** Normal Worker ****                    
                     // A worker with an input and output queue.
-                    final byte[] output = codec.serialise(responseMessage);
-                    workerQueue.publish(taskInformation, output, queue, Collections.emptyMap(), true);
-                    stats.getOutputSizes().update(output.length);
+                    workerQueue.publish(taskInformation, responseMessage, queue, Collections.emptyMap(), true);
+                    stats.getOutputSizes().update(responseMessage.getTaskData().length);
```
dermot-hardy: Same question here: Are we sure `.getTaskData()` will never be `null`?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/InvalidDeliveryException.java
```diff
@@ -0,0 +1,40 @@
...
+package com.github.workerframework.queues.rabbit;
+
+public class InvalidDeliveryException extends Exception {
+    private static final long serialVersionUID = 1L;
+    private final long messageId;
+
+    /**
+     * Create a new InvalidDeliveryException with the specified message.
+     *
+     * @param message the message to include in the exception
```
dermot-hardy: Javadoc for `message` but not `messageId`.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/InvalidDeliveryException.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.workerframework.queues.rabbit;
+
+public class InvalidDeliveryException extends Exception {
```
dermot-hardy: Does this need to be `public`?

```suggestion
class InvalidDeliveryException extends Exception {
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/TransientDeliveryException.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.workerframework.queues.rabbit;
+
+public class TransientDeliveryException extends Exception {
```
dermot-hardy: Does this need to be public?

```suggestion
class TransientDeliveryException extends Exception {
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/TransientDeliveryException.java
```diff
@@ -0,0 +1,35 @@
...
+    private static final long serialVersionUID = 1L;
+    private final long messageId;
+
+    /**
+     * Create a new TransientDeliveryException with the specified message.
+     *
+     * @param message the message to include in the exception
```
dermot-hardy: Missing params in the Javadoc (or just remove the Javadoc if this class doesn't need to be public).

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitWorkerQueueConfiguration.java
```diff
@@ -87,6 +87,23 @@ public class RabbitWorkerQueueConfiguration
...
+    /**
+     * The threshold at which message payloads will be offloaded before publishing to RabbitMQ.
+     */
+    @Min(1)
+    private int payloadOffloadingThreshold = 16777216;
+
+    /**
+     * The datastore directory to use for offloading payloads.
+     */
+    @NotNull
```
dermot-hardy: This could force worker updates to be required.  Can we not just accept `null` and have a default in the code for when that is the case?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitWorkerQueue.java
```diff
@@ -72,15 +81,27 @@ public final class RabbitWorkerQueue implements ManagedWorkerQueue
...
+    public RabbitWorkerQueue(
+        RabbitWorkerQueueConfiguration config,
+        int maxTasks,
+        final String invalidQueue,
+        final ManagedDataStore dataStore,
+        final Codec codec)
     {
         this.config = Objects.requireNonNull(config);
         this.maxTasks = maxTasks;
+        this.invalidQueue = invalidQueue;
```
dermot-hardy: The two other new parameters being added have `Objects.requireNonNull()` but this one doesn't.  Does that imply that we are ok with `null` being accepted here?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitWorkerQueue.java
```diff
@@ -140,11 +177,54 @@ public void publish(TaskInformation taskInformation, byte[] taskMessage, String
...
+        publisherQueue.add(new WorkerPublishQueueEvent(serializedTaskMessage, targetQueue, rabbitTaskInformation, publishHeaders));
     }
-    
+
+    private String getStoragePath(final String routingKey, final String trackingJobTaskId)
+    {
+        final StringBuilder path = new StringBuilder(Paths.get(config.getPayloadOffloadingDirectory(), routingKey).toString());
+        final Matcher matcher = JOB_TASK_ID_PATTERN.matcher(trackingJobTaskId);
+        if (matcher.find()) {
+            path.append("/" + matcher.group(1).replace(":", "/"));
```
dermot-hardy:
```suggestion
            path.append('/').append(matcher.group(1).replace(":", "/"));
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitWorkerQueue.java
```diff
@@ -140,11 +177,54 @@ public void publish(TaskInformation taskInformation, byte[] taskMessage, String
...
+            if (matcher.group(2) != null && !matcher.group(2).isEmpty()) {
+                path.append("/" + matcher.group(2));
```
dermot-hardy:
```suggestion
                path.append('/').append(matcher.group(2));
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -34,86 +42,231 @@
...
         this.metrics = Objects.requireNonNull(metrics);
         this.consumerEventQueue = Objects.requireNonNull(queue);
         this.channel = Objects.requireNonNull(ch);
         this.publisherEventQueue = Objects.requireNonNull(pubQueue);
         this.retryRoutingKey = Objects.requireNonNull(retryKey);
         this.retryLimit = retryLimit;
+        this.invalidRoutingKey = invalidKey;
+        this.dataStore = Objects.requireNonNull(dataStore);
+        this.codec = Objects.requireNonNull(codec);
+        this.disconnectCallback = disconnectCallback;
```
dermot-hardy: Using `Objects.requireNonNull()` with `dataStore` and `codec`, but not with `invalidKey` and `disconnectCallback` makes it look like `null` is acceptable for these, which I assume is not the case?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -34,86 +42,231 @@
...
+        }
+        catch (final InvalidDeliveryException ex) {
+            LOG.error("Invalid delivery for message id {}: {}", ex.getMessageId(), ex.getMessage());
+
+            final RabbitTaskInformation taskInformation = new RabbitTaskInformation(String.valueOf(inboundMessageId), true);
+            taskInformation.incrementResponseCount(true);
+            final var publishHeaders = new HashMap<String, Object>();
+
+            if(ex.getCause() != null && ex.getCause() instanceof ReferenceNotFoundException) {
+                publishHeaders.put(RabbitHeaders.RABBIT_HEADER_CAF_WORKER_REJECTED, REJECTED_REASON_PAYLOAD_OFFLOADING_TASKMESSAGE_DATASTOREE_REFERENCE_NOT_FOUND_ERROR);
```
dermot-hardy:
```suggestion
                publishHeaders.put(RabbitHeaders.RABBIT_HEADER_CAF_WORKER_REJECTED,
                                   REJECTED_REASON_PAYLOAD_OFFLOADING_TASKMESSAGE_DATASTOREE_REFERENCE_NOT_FOUND_ERROR);
```

dermot-hardy (approved):

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: Didn't we encounter this infinite loop early days? Was that caused by the null byte array deserialization?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -34,86 +42,231 @@
...
  * is not marked 'redelivered'. Redelivered messages are republished to the retry queue with an incremented retry count. Redelivered
  * messages that have exceeded the retry count are republished to the rejected queue.
  */
 public class WorkerQueueConsumerImpl implements QueueConsumer
 {
-    public static final String REJECTED_REASON_TASKMESSAGE = "TASKMESSAGE_INVALID";
+    public static final String REJECTED_REASON_TASKMESSAGE_INVALID = "TASKMESSAGE_INVALID";
+    private static final String REJECTED_REASON_PAYLOAD_OFFLOADING_TASKMESSAGE_DATASTOREE_REFERENCE_NOT_FOUND_ERROR = 
+            "TASKMESSAGE_DATASTORE_REFERENCE_NOT_FOUND_ERROR";
+
```
DavidGerald-Milligan: typo? REJECTED_REASON_PAYLOAD_OFFLOADING_TASKMESSAGE_**DATASTOREE**_REFERENCE_NOT_FOUND_ERROR

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
DavidGerald-Milligan: Actually I recall now, it was the null byte array when the header was missing, we fixed that in the message prioritization when we copied in the missing headers.

worker-test/src/test/java/com/github/workerframework/workertest/WorkerTestBase.java
```diff
@@ -51,8 +51,8 @@ public class WorkerTestBase {
     private static final String CAF_RABBITMQ_PORT = "CAF_RABBITMQ_PORT";
     private static final String CAF_RABBITMQ_USERNAME = "CAF_RABBITMQ_USERNAME";
     private static final String CAF_RABBITMQ_PASSWORD = "CAF_RABBITMQ_PASSWORD";
-    public static final String WEBDAV_URL = System.getProperty("WEBDAV_URL", "http://localhost:9090/webdav");
-    private static final Codec codec = new JsonCodec();
+    protected static final String WEBDAV_URL = System.getProperty("WEBDAV_URL", "http://localhost:9090/webdav");
```
DavidGerald-Milligan: I dont see any use of this outside WorkerTestBase

worker-test/src/test/java/com/github/workerframework/workertest/PayloadOffloadingIT.java
```diff
@@ -79,6 +79,29 @@ public void checkOffloadedPayloadIsConsumedAndDeletedOnAck() throws Exception {
...
+            final Channel channel = prepareChannel(connection, WORKER_IN, WORKER_OUT)) {
+
+            //  Now we can send a message which expects to find the setupPayloadOffloadStorageRef.
+            final Map<String, Object> headers = new HashMap<>();
+            headers.put(RABBIT_HEADER_CAF_PAYLOAD_OFFLOADING_STORAGE_REF, UUID.randomUUID().toString());
+            publish(channel, codec.serialise(taskMessage), headers, WORKER_IN);
+
+            final TestWorkerQueueConsumer outboundConsumer = new TestWorkerQueueConsumer();
+            consume(channel, outboundConsumer, WORKER_OUT);
+            Assert.assertEquals("", new String(outboundConsumer.getLastDeliveredBody(), StandardCharsets.UTF_8));
```
DavidGerald-Milligan: Flip arguments, but shouldn't getLastDeliveredBody be returning the TaskMessage with an empty Taskdata byteArray?

worker-api/src/main/java/com/github/workerframework/api/TaskMessage.java
```diff
@@ -53,7 +53,6 @@ public final class TaskMessage
     /**
      * The serialised data of the task-specific message.
      */
-    @NotNull
     private byte[] taskData;
```
DavidGerald-Milligan, line 56: The ctor has a null check: _this.taskData = Objects.requireNonNull(taskData);_

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -175,21 +339,43 @@ private void processReject(long id, boolean requeue)
...
+                    dataStore.store(dataStore.retrieve(taskMessageStorageRefOpt.get()), taskMessageStorageRefOpt.get().replace(deliveryQueue, retryRoutingKey));
+                } 
+                catch (final ReferenceNotFoundException e) {
+                    throw new InvalidDeliveryException("Original reference not found when relocating TaskData", inboundMessageId, e);
+                }
+                catch (final DataStoreException e) {
+                    LOG.error("Failed to relocate offloaded payload for message id {} from {} to {}",
+                            inboundMessageId, deliveryQueue, retryRoutingKey, e);
+                    disconnectCallback.run();
+                }
```
DavidGerald-Milligan, line 454: Does retrieve only throw the ReferenceNotFoundException?
Don't we need to update the publish headers here too? And delete the now relocated data?

andyreidz:
@DavidGerald-Milligan Could you have another look at the recent changes?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -166,10 +166,11 @@ public void processDelivery(Delivery delivery)
                 publishHeaders.put(RABBIT_HEADER_CAF_WORKER_INVALID, ex);
             }
 
-            publisherEventQueue.add(new WorkerPublishQueueEvent(deliveryMessageData, invalidRoutingKey, taskInformation, deliveryHeaders));
+            publisherEventQueue.add(new WorkerPublishQueueEvent(deliveryMessageData, invalidRoutingKey, taskInformation, publishHeaders));
         } catch (final TransientDeliveryException e) {
             LOG.warn("Transient error processing message id {}, disconnecting.", inboundMessageId, e);
             offloadedPayloadsToDelete.remove(inboundMessageId);
+            //Disconnect the channel to allow for a reconnect when the HealthCheck passes.
             disconnectCallback.run();
```
DavidGerald-Milligan, line 183: Only vaguely related but I see that the readme states  " _- If a transient error occurs in processing, throw a TaskRejectedException_" does that need updated?

worker-test/src/test/java/com/github/workerframework/workertest/PayloadOffloadingIT.java
```diff
@@ -161,9 +183,9 @@ public void checkOffloadedPayloadIsDeletedOnTerminalWorker() throws Exception {
         Assert.assertTrue(readWebDAVFile.isPresent(), "The file should be present in the datastore");
 
         try(final Connection connection = connectionFactory.newConnection();
-            final Channel channel = prepareChannel(connection, TERMINAL_WORKER_IN, TERMINAL_WORKER_OUT, 
-                    TERMINAL_WORKER_INVALID)) {
-
+            final Channel channel = prepareChannel(connection)) {
+            createQueues(channel, TERMINAL_WORKER_IN, TERMINAL_WORKER_OUT);
+
```
DavidGerald-Milligan: whitespace

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
andyreidz: It's still correct in that when the worker is doing it work (in doWork) a TaskRejectedException will be thrown but this is before the worker is given the task so handled differently.

dermot-hardy:
What were all the changes made here after the review approvals?  Were they trivial?

DavidGerald-Milligan:
> What were all the changes made here after the review approvals? Were they trivial?

The main update was the [reporting PR](https://github.com/WorkerFramework/worker-framework/pull/227) which was merged with this, subsequent minor updates were made following testing of that.  

The only other noteable update:
We had to [pass the workerName](https://github.com/WorkerFramework/worker-framework/pull/214/commits/d795cc7ac1233e525ed0154d38921fd554c6f606) to enable reporting, because an issue was discovered when integrating with workers that implemented an AbstractWorkerConfiguration that dis not have a WorkerConfiguration.

Anything else was trivial.

All changes were however discussed at length with, and approved by @andyreidz

andyreidz:
@dermot-hardy I'm not sure why the reviews from June weren't marked as stale by GitHub. David is right that the main change since the last review was to correct the reporting of invalid tasks to job service but you haven't seen those changes and I very much appreciate your input to them. Is it practical for you to comment on this merged PR and then we decide next step?

worker-core/pom.xml
```diff
@@ -48,6 +48,10 @@
             <groupId>com.github.workerframework</groupId>
             <artifactId>worker-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.workerframework</groupId>
+            <artifactId>worker-caf</artifactId>
+        </dependency>
```
dermot-hardy, line 54: `worker-core` taking a dependency on `worker-caf` is very unexpected given that `worker-caf` is meant to be just some abstract classes to help implement workers.  What is the justification for this?

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -370,4 +372,25 @@ private static HealthCheckConfiguration createHealthCheckConfiguration(
...
         return healthCheckConfiguration;
     }
+
+    /**
+     * Some instances of WorkerFactory are implementations of AbstractWorkerFactory<C,T> where C
+     * sometimes implements WorkerConfiguration.  In those cases we take the workerName from
+     * the WorkerConfiguration, otherwise we take the workerName from AbstractWorkerFactory.
+     *
+     * Note there are occasions where the workerName is defined differently in AbstractWorkerFactory
+     * to that in WorkerConfiguration. e.g. worker-batch.
```
dermot-hardy, line 382: This seems to be the answer to my previous question but this seems like a bad idea.  I really don't like `worker-core` having special logic for `AbstractWorkerFactory`, which is just a barely used (badly written) convenience class.  The `AbstractWorkerFactory.getWorkerName()` is just badly named &mdash; it's not even the name of the worker.

worker-caf/src/main/java/com/github/workerframework/caf/AbstractWorkerFactory.java
```diff
@@ -159,7 +159,7 @@ private T verifyWorkerTask(
     /**
      * @return the name of the worker to use when checking the task type
      */
-    protected abstract String getWorkerName();
+    public abstract String getWorkerName();
```
dermot-hardy, line 162: I don't like this change.  Discussed in previous comments.

worker-queue-rabbit/pom.xml
```diff
@@ -48,6 +48,10 @@
             <groupId>com.github.workerframework</groupId>
             <artifactId>worker-configs</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.workerframework</groupId>
+            <artifactId>worker-tracking-report</artifactId>
+        </dependency>
```
dermot-hardy, line 54: Why does `worker-queue-rabbit` now have a dependency on `worker-tracking-report`?  These seem like they should be very unrelated things.

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/RabbitWorkerQueue.java
```diff
@@ -107,13 +131,31 @@ public void start(TaskCallback callback)
...
+            WorkerPublisherImpl publisherImpl = new WorkerPublisherImpl(
+                outgoingChannel,
+                metrics,
+                consumerQueue,
+                confirmListener
+            );
             publisher = new EventPoller<>(2, publisherQueue, publisherImpl);
             declareWorkerQueue(incomingChannel, config.getInputQueue());
             declareWorkerQueue(outgoingChannel, config.getRetryQueue());
+            declareWorkerQueue(outgoingChannel, invalidQueue);
```
dermot-hardy, line 158: Does this mean that we're creating lots of invalid queues that may never be used?  (I'm not sure if that's an issue if it does mean that.  It might not be.)

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -63,57 +104,249 @@ public WorkerQueueConsumerImpl(TaskCallback callback, RabbitMetricsReporter metr
...
+            throws ReferenceNotFoundException, TransientDeliveryException
+    {
+        try (final var inputStream = dataStore.retrieve(taskMessageStorageRef)) {
+            final var taskData = inputStream.readAllBytes();
+            offloadedPayloadsToDelete.put(inboundMessageId, taskMessageStorageRef);
+            return taskData;
+        } catch (final IOException | DataStoreException ex) {
+            if (ex instanceof ReferenceNotFoundException) {
+                throw (ReferenceNotFoundException)ex;
+            }
```
dermot-hardy, line 230: Why is the `ReferenceNotFoundException` not just being caught first?

```suggestion
        } catch (final ReferenceNotFoundException ex) {
            throw ex;
        } catch (final IOException | DataStoreException ex) {
```

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -63,57 +104,249 @@ public WorkerQueueConsumerImpl(TaskCallback callback, RabbitMetricsReporter metr
...
+
+            publisherEventQueue.add(new WorkerPublishQueueEvent(deliveryMessageData, invalidRoutingKey, taskInformation, publishHeaders));
+
+            if (deliveredTaskMessageOpt.isPresent()) {
+                final var taskMessage = deliveredTaskMessageOpt.get();
+                if(taskMessage.getTracking() != null) {
+                    sendFailureTrackingReport(taskMessage, exceptionMesssage, taskInformation);
+                }
+            }
+        } catch (CodecException e) {
```
dermot-hardy, line 258:
```suggestion
        } catch (final CodecException e) {
```

PR #223, opened 2025-05-14
US1027066: Enabling integration tests in jenkins build

DavidGerald-Milligan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1027066


worker-test/pom.xml
```diff
@@ -88,13 +88,6 @@
 
     <build>
         <plugins>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
-                </configuration>
-            </plugin>
```
DavidGerald-Milligan, line 97: This was added to stop ShutdownDeveloperTest from running.  Since its been renamed ShutdownDeveloperIT its no longer needed.

worker-test/src/test/java/com/github/workerframework/workertest/PoisonMessageIT.java
```diff
@@ -69,6 +68,40 @@ public void getWorkerNameInPoisonMessageTest() throws Exception {
                     .build();
 
             channel.basicPublish("", WORKER_IN, properties, codec.serialise(requestTaskMessage));
+            
+            // Test worker gets killed, the message SHOULD get redelivered, but it is not
+            // so we will consume and nack the message, forcing a redelivery which
+            // will be handled as a poison message.
+            nackMessage(channel);
```
DavidGerald-Milligan, line 71: If test worker gets killed, the message SHOULD get redelivered, but it is not so we will consume and nack the message, forcing a redelivery which will be handled as a poison message.

worker-test/pom.xml
```diff
@@ -363,50 +359,6 @@
...
-                                </log>
-                                <wait>
-                                    <http>
-                                        <url>http://${docker.host.address}:${worker.testadminport2}</url>
-                                    </http>
-                                    <time>120000</time>
-                                    <shutdown>500</shutdown>
-                                </wait>
-                            </run>
-                        </image>
```
DavidGerald-Milligan, line 409: Not relying on killing worker any more so no need for second worker

DavidGerald-Milligan:
Closing this, superceded by changes in https://github.com/WorkerFramework/worker-framework/pull/225

PR #225, opened 2025-05-19
1009117 Refactored for readability

andyreidz:


DavidGerald-Milligan (approved):

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -83,105 +91,132 @@ public WorkerQueueConsumerImpl(TaskCallback callback, RabbitMetricsReporter metr
...
                 if (retries < retryLimit) {
-                    //Republish the delivery with a header recording the incremented number of retries.
-                    //Classic queues do not record delivery count, so we republish the message with an incremented
-                    //retry count. This allows us to track the number of attempts to process the message.
-                    republishClassicRedelivery(delivery, retries, taskMessageStorageRefOpt, taskMessage.getTracking());
-                    return;
+                    republishClassicRedelivery(
+                        delivery, retries, trackingInfo
+                    );
+                    return PoisonMessageStatus.CLASSIC_AND_REPUBLISHED;
```
DavidGerald-Milligan, line 212: @andyreidz

DavidGerald-Milligan, line 215: @andyreidz With the update to how we clean up the datastore, will this result in an ack, and therefore a delete, of the stored inbound message? If so should we pass the byte array read from the store rather than the delivery.getMessagedata()?

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -246,33 +265,34 @@ public void processAck(long tag)
             metrics.incremementErrors();
             consumerEventQueue.add(new ConsumerAckEvent(tag));
         }
+
```
DavidGerald-Milligan, line 272: @andyreidz If the ack fails above, should we be falling through here and deleting the message from the datastore?

DavidGerald-Milligan (approved):

DavidGerald-Milligan:
@andyreidz I have reviewed again and made some minor updates to tests.  I deployed update-entities and classification workers employing this build, and the updated message-distribution worker where we pass the original headers on larry and re-tested the 'mark false positive' function.  The the headers are passed through to the workers and the issue of the message looping is no longer present.  The flow through the workers is good, and I checked the datastore on larry and the stored files are cleaned up as expected, though I do notice that the directories created in storing the taskMessage are still present, albeit empty.

PR #226, opened 2025-06-13
1026274: Worker Framework Shutdown hanging

andyreidz:
Deadlock was caused by the use of System.exit in the test worker, using this opportunity to simplify shutdown logic and use a non-configurable timeout.

Update the shutdownIncoming method to make the shutdown permanent so GatedHealthProvider does not trigger a reconnect.

```
[2025-08-05 13:27:46.140Z #110.028 INFO  -            -   ] c.g.w.core.WorkerApplication: Worker stop requested.
[2025-08-05 13:27:46.145Z #110.028 INFO  -            -   ] c.g.w.core.WorkerApplication: Allowing 10 backlog tasks to complete, 1 currently active.
[2025-08-05 13:28:01.150Z #110.028 INFO  -            -   ] c.g.w.core.WorkerApplication: Allowing 7 backlog tasks to complete, 1 currently active.
...
[2025-08-05 13:28:31.160Z #110.028 INFO  -            -   ] c.g.w.core.WorkerApplication: Allowing 1 backlog tasks to complete, 1 currently active.
[2025-08-05 13:28:46.197Z #110.028 INFO  -            -   ] c.g.w.core.WorkerApplication: Worker stopped.
```

### DIFF
--- a/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
@@ -150,6 +150,11 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
     {
         return workQueue.size() + backupThreadPool.getBacklogSize();
     }
+    
+    @Override 
+    public int getActiveCount() {
+        return backupThreadPool.getActiveCount();
+    }
 
     @Override
     public void submitWorkerTask(final WorkerTaskImpl workerTask)
--- a/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
@@ -37,7 +37,7 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
     private final StreamingWorkerThreadPool backupThreadPool;
 
     private volatile boolean isActive;
-    private volatile AtomicInteger activeThreads = new AtomicInteger(0);
+    private final AtomicInteger activeThreads = new AtomicInteger(0);
 
     public BulkWorkerThreadPool(
         final WorkerFactory workerFactory,
@@ -88,7 +88,7 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
                 = new BulkWorkerTaskProvider(task, workQueue);
 
             try {
-                activeThreads.addAndGet(1);
+                activeThreads.incrementAndGet();
                 bulkWorker.processTasks(taskProvider);
             } catch (final RuntimeException ex) {
                 LOG.warn("Bulk Worker threw unhandled exception", ex);
--- a/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -36,6 +37,7 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
     private final StreamingWorkerThreadPool backupThreadPool;
 
     private volatile boolean isActive;
+    private volatile AtomicInteger activeThreads = new AtomicInteger(0);
 
     public BulkWorkerThreadPool(
         final WorkerFactory workerFactory,
@@ -86,10 +88,12 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
                 = new BulkWorkerTaskProvider(task, workQueue);
 
             try {
+                activeThreads.addAndGet(1);
                 bulkWorker.processTasks(taskProvider);
             } catch (final RuntimeException ex) {
                 LOG.warn("Bulk Worker threw unhandled exception", ex);
             } finally {
+                activeThreads.decrementAndGet();
                 // Re-submit the first task if it has not been consumed
                 // NB: It's really faulty Worker logic to not consume at least
                 // the one task.
@@ -152,8 +156,8 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
     }
     
     @Override 
-    public int getActiveCount() {
-        return backupThreadPool.getActiveCount();
+    public int getApproxActiveCount() {
+        return activeThreads.get() + backupThreadPool.getApproxActiveCount();
     }
 
     @Override
--- a/worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
@@ -78,7 +78,7 @@ final class StreamingWorkerThreadPool implements WorkerThreadPool
     }
     
     @Override
-    public int getActiveCount() {
+    public int getApproxActiveCount() {
         return threadPoolExecutor.getActiveCount();
     }
 
--- a/worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
@@ -76,6 +76,11 @@ final class StreamingWorkerThreadPool implements WorkerThreadPool
     {
         return workQueue.size();
     }
+    
+    @Override
+    public int getActiveCount() {
+        return threadPoolExecutor.getActiveCount();
+    }
 
     /**
      * Execute the specified task at some point in the future
--- a/worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
+++ b/worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
@@ -151,10 +151,12 @@ public final class WorkerApplication extends Application<WorkerConfiguration>
 
                 final long startTime = System.currentTimeMillis();
                 
-                while(wtp.getBacklogSize() > 0 && System.currentTimeMillis() - startTime < SHUTDOWN_DURATION) {
+                int backlogSize = wtp.getBacklogSize();
+                while(backlogSize > 0 && System.currentTimeMillis() - startTime < SHUTDOWN_DURATION) {
                     try {
-                        LOG.info("Allowing {} backlog tasks to complete, {} currently active.", wtp.getBacklogSize(), wtp.getApproxActiveCount());
+                        LOG.debug("Allowing {} backlog tasks to complete, {} currently active.", backlogSize, wtp.getApproxActiveCount());
                         Thread.sleep(SHUTDOWN_LOG_INTERVAL);
+                        backlogSize = wtp.getBacklogSize();
                     } catch (final InterruptedException e) {
                         Thread.currentThread().interrupt();
                         break;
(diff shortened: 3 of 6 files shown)


worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -145,23 +145,11 @@ public void start() {
...
-                        LOG.trace("Awaiting the Worker Thread Pool to become idle, {} tasks in the backlog.", 
-                                wtp.getBacklogSize());
-                        Thread.sleep(1000);
-                    } catch (final InterruptedException e) {
-                        Thread.currentThread().interrupt();
-                        throw new RuntimeException(e);
-                    }
-                }
-                LOG.trace("Worker Thread Pool is idle.");
                 wtp.shutdown();
```
andyreidz, line 165: The above isn't needed as wtp.shutdown allows existing tasks to complete.

andyreidz: BulkWorkerThreadPool should be changed to backupThreadPool.shutdown(), backupThreadPool.awaitTermination() and then interrupt the worker threads.

```
    @Override
    public void shutdown()
    {
        isActive = false;

        for (BulkWorkerThread workerThread : bulkWorkerThreads) {
            workerThread.interrupt();
        }

        backupThreadPool.shutdown();
    }
```

andyreidz:
More thought is required for BulkWorkerThreadPool

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
dermot-hardy: The code being removed was added for #175.

andyreidz: Yes the check wtp.isIdle() was flawed because ultimately it calls ThreadPoolExecutor.getActiveCount to determine if it is active, the JavaDoc for that method says "Returns the approximate number of threads that are actively executing tasks."

What remains from #175 is the change from using Runtime.getRuntime().addShutdownHook to using the dropwizard application lifecycles hooks environment.lifecycle().manage(new Managed().

The wtp.awaitTermination(10_000, TimeUnit.MILLISECONDS); also changed to be 5 minutes.

worker-test/src/test/java/com/github/workerframework/workertest/ShutdownDeveloperTest.java
```diff
@@ -45,7 +45,7 @@ public class ShutdownDeveloperTest extends TestWorkerTestBase {
     public void shutdownTest() throws IOException, TimeoutException, CodecException {
 
         // Usage instructions
-        // Comment out the iages for test worker 2 and 3 in this module's pom.xml
+        // Comment out the images for worker-test-2 and worker-test-no-valid-cert in this module's pom.xml
```
andyreidz, line 48: This can be used to manually test what happens when the container is stopped.

worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
```diff
@@ -150,6 +150,11 @@ public int getBacklogSize()
     {
         return workQueue.size() + backupThreadPool.getBacklogSize();
     }
+    
+    @Override 
+    public int getActiveCount() {
+        return backupThreadPool.getActiveCount();
```
dermot-hardy: This implementation seems somewhat dubious, given that the backup thread pool is not normally used &mdash; it only comes into play when there is an error and the bulk worker falls back to trying to process documents one at a time.  For a bulk worker normally it is the `bulkWorkerThreads` that are used.

worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
```diff
@@ -76,6 +76,11 @@ public int getBacklogSize()
     {
         return workQueue.size();
     }
+    
+    @Override
+    public int getActiveCount() {
+        return threadPoolExecutor.getActiveCount();
```
dermot-hardy, line 82: It's interesting that the JavaDoc says *approximate* number of threads:

image

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -78,6 +78,8 @@ public final class WorkerApplication extends Application<WorkerConfiguration>
 {
     private final long startTime = System.currentTimeMillis();
     private static final Logger LOG = LoggerFactory.getLogger(WorkerApplication.class);
+    private static final long SHUTDOWN_DURATION = 5 * 60 * 1000; // 5 minutes in milliseconds
+    public static final int SHUTDOWN_LOG_INTERVAL = 15_000; // 15 seconds in milliseconds
```
dermot-hardy: I assume this is not meant to be `public`?

```suggestion
    private static final int SHUTDOWN_LOG_INTERVAL = 15_000; // 15 seconds in milliseconds
```

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -143,25 +145,27 @@ public void start() {
...
-                        //The grace period will expire and the process killed so no need for time limit here
-                        LOG.trace("Awaiting the Worker Thread Pool to become idle, {} tasks in the backlog.", 
-                                wtp.getBacklogSize());
-                        Thread.sleep(1000);
+                        LOG.info("Allowing {} backlog tasks to complete, {} currently active.", wtp.getBacklogSize(), wtp.getActiveCount());
+                        Thread.sleep(SHUTDOWN_LOG_INTERVAL); // 15 seconds
                     } catch (final InterruptedException e) {
                         Thread.currentThread().interrupt();
-                        throw new RuntimeException(e);
+                        System.out.println("Logger thread interrupted, exiting...");
```
dermot-hardy: Was this line meant to be left in?  If so should it not use the logger rather than `System.out.println()`?  Also what does the message mean by "logger thread"?

worker-core/src/main/java/com/github/workerframework/core/WorkerThreadPool.java
```diff
@@ -61,6 +61,8 @@ void awaitTermination(long timeout, TimeUnit unit)
     boolean isIdle();
 
     int getBacklogSize();
+    
+    int getActiveCount();
```
dermot-hardy: I see now that this is only used for logging.  Maybe it would be better to call it `getApproxActiveCount()` just to emphasize that the result does not impact functionality?

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -143,25 +145,27 @@ public void start() {
...
+                final long startTime = System.currentTimeMillis();
+                
+                while(wtp.getBacklogSize() > 0 && System.currentTimeMillis() - startTime < SHUTDOWN_DURATION) {
                     try {
-                        //The grace period will expire and the process killed so no need for time limit here
-                        LOG.trace("Awaiting the Worker Thread Pool to become idle, {} tasks in the backlog.", 
-                                wtp.getBacklogSize());
-                        Thread.sleep(1000);
+                        LOG.info("Allowing {} backlog tasks to complete, {} currently active.", wtp.getBacklogSize(), wtp.getActiveCount());
+                        Thread.sleep(SHUTDOWN_LOG_INTERVAL); // 15 seconds
```
dermot-hardy: Won't this cause normal shutdowns to take much longer?  Why do we want to move this from 1 second to 15 seconds?

worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
andyreidz: Yes, this value is only used for logging so I was happy to use the approximate value.

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
andyreidz: We won't enter the while loop if there's no backlog and we don't want to log too much so normal shutdown will be ok?

worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
andyreidz: Updated to track active threads.

worker-core/src/main/java/com/github/workerframework/core/StreamingWorkerThreadPool.java
andyreidz: Renamed as suggested below

worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
```diff
@@ -86,10 +88,12 @@ private void execute()
                 = new BulkWorkerTaskProvider(task, workQueue);
 
             try {
+                activeThreads.addAndGet(1);
```
dermot-hardy: I know it's the same but `increment` would be more normal than `add(1)`.

```suggestion
                activeThreads.incrementAndGet();
```

worker-core/src/main/java/com/github/workerframework/core/BulkWorkerThreadPool.java
```diff
@@ -36,6 +37,7 @@ final class BulkWorkerThreadPool implements WorkerThreadPool
     private final StreamingWorkerThreadPool backupThreadPool;
 
     private volatile boolean isActive;
+    private volatile AtomicInteger activeThreads = new AtomicInteger(0);
```
dermot-hardy: This shouldn't be `volatile`, and it should be `final`.

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
dermot-hardy: A backlog being present will be a regular event.  To avoid logging too much could we just put the log statement outside the loop, or downgrade it to debug, or add some extra logic so that it is not logged in every iteration.

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -143,25 +145,26 @@ public void start() {
...
+
+                final long startTime = System.currentTimeMillis();
+                
+                while(wtp.getBacklogSize() > 0 && System.currentTimeMillis() - startTime < SHUTDOWN_DURATION) {
                     try {
-                        //The grace period will expire and the process killed so no need for time limit here
-                        LOG.trace("Awaiting the Worker Thread Pool to become idle, {} tasks in the backlog.", 
-                                wtp.getBacklogSize());
-                        Thread.sleep(1000);
+                        LOG.info("Allowing {} backlog tasks to complete, {} currently active.", wtp.getBacklogSize(), wtp.getApproxActiveCount());
```
dermot-hardy: The backlog size may have changed between the decision to enter the loop and this logging statement.  Should it be put into a variable to log the value that made it come into the loop, or are we ok with it logging an updated value?

worker-core/src/main/java/com/github/workerframework/core/WorkerApplication.java
```diff
@@ -143,25 +145,26 @@ public void start() {
...
             public void stop() {
-                LOG.info("Worker stop requested, allowing in-progress tasks to complete.");
+                LOG.info("Worker stop requested.");
+                
                 workerQueue.shutdownIncoming();
-                while (!wtp.isIdle()) {
+
+                final long startTime = System.currentTimeMillis();
+                
+                while(wtp.getBacklogSize() > 0 && System.currentTimeMillis() - startTime < SHUTDOWN_DURATION) {
```
dermot-hardy: Are we sure about this strategy of trying to clear the backlog during shutdown?  Trying to complete in-progress tasks makes sense but if I'm understanding this right it is intentionally allowing new tasks to begin that haven't been passed to the worker code yet.  Wouldn't we be better to clear these backlog queues and let the next worker instance pick them up?

andyreidz: The prefetched message have their delivery count increased and this goes towards the poison message identification. 
I take your point that starting more messages increases the chances they get terminated after the timeout.
Let's aim to talk it through at some stage.

PR #237, opened 2025-09-12
US1009117: delete storage dir

DavidGerald-Milligan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1009117


worker-api/src/main/java/com/github/workerframework/api/DataStore.java
```diff
@@ -32,6 +32,18 @@ public interface DataStore
...
+    /**
+     * Delete asset identified by reference
+     *
+     * @param reference a complete reference to be interpreted by the DataStore implementation
+     * @throws DataStoreException if data store cannot service the request
+     */
+    default void deleteTree(final String reference)
+            throws DataStoreException
+    {
+        delete(reference);
```
andyreidz, line 47: I'm not sure about adding this to the DataStore interface, we'd need to think about how different implementations would behave as the concept of a directory tree is not something all implementations would have.

andyreidz (changes requested):
We can discuss as a group.

dermot-hardy (approved):

worker-queue-rabbit/src/main/java/com/github/workerframework/queues/rabbit/WorkerQueueConsumerImpl.java
```diff
@@ -383,7 +383,7 @@ public void processAck(long tag)
         final String datastorePayloadReference = offloadedPayloadsToDelete.remove(tag);
         if (datastorePayloadReference != null) {
             try {
-                dataStore.delete(datastorePayloadReference);
+                dataStore.delete(datastorePayloadReference, true);
             } catch (final DataStoreException e) {
                 LOG.warn("Couldn't delete offloaded payload '{}' for delivery tag '{}' from datastore message.",
```
andyreidz, line 388: We'll need a performance test where we achieve a very high message throughput to verify that repeated exceptions (when the parent directories are not empty) don't hurt performance.

andyreidz (approved):

PR #247, opened 2026-01-08
US1055357: stop deleting queues directory on partial ref cleanup

DavidGerald-Milligan:
https://internal.almoctane.com/ui/entity-navigation?p=131002/6001&entityType=work_item&id=1055357

### DIFF
--- a/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
+++ b/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
@@ -15,6 +15,7 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
+    noDeleteDirs: JSON.parse(getenv("CAF_WORKER_DATASTORE_NO_DELETE_DIRS") || '["queues"]'),
     dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS") || undefined,
     outputBufferSize: getenv("CAF_WORKER_DATASTORE_OUTPUT_BUFFER_SIZE")
             || getenv("CAF_WORKER_DATASTORE_BUFFER_SIZE") || undefined
--- a/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
+++ b/worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
@@ -15,7 +15,6 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
-    noDeleteDirs: JSON.parse(getenv("CAF_WORKER_DATASTORE_NO_DELETE_DIRS") || '["queues"]'),
     dataDirHealthcheckTimeoutSeconds: getenv("CAF_WORKER_DATASTORE_HEALTHCHECK_TIMEOUT_SECONDS") || undefined,
     outputBufferSize: getenv("CAF_WORKER_DATASTORE_OUTPUT_BUFFER_SIZE")
             || getenv("CAF_WORKER_DATASTORE_BUFFER_SIZE") || undefined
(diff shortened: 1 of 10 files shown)


worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
```diff
@@ -15,6 +15,7 @@
  */
 ({
     dataDir: getenv("CAF_WORKER_DATASTORE_PATH") || "/mnt/caf-datastore-root",
+    noDeleteDirs: JSON.parse(getenv("CAF_WORKER_DATASTORE_NO_DELETE_DIRS") || '["queues"]'),
```
michael-bryson: Just wondering if we really want to default this to "queues" as that is quite specific for our use case and this is an open source project?

DavidGerald-Milligan: Its the same as we do for [RabbitWorkerQueueConfiguration](https://github.com/WorkerFramework/worker-framework/blob/develop/worker-default-configs/config/cfg~caf~worker~RabbitWorkerQueueConfiguration.js)

michael-bryson: Sure but that is specific to rabbit, this one is not. I don't mind either way just wasn't sure about defaulting it to a value related to rabbit and our use case.

rorytorneymf (approved):

worker-default-configs/config/cfg~caf~worker~FileSystemDataStoreConfiguration.js
DavidGerald-Milligan: @dermot-hardy have you any thoughts on @michael-bryson's point. If this is an issue we could add the default to application-deploy.

dermot-hardy: I think we do need to do something about it.  If the operator changes the `CAF_WORKER_PAYLOAD_OFFLOADING_DIRECTORY` setting they shouldn't also have to change `CAF_WORKER_DATASTORE_NO_DELETE_DIRS` to keep it in-sync.  It should change the defaults for both.

DavidGerald-Milligan: I've removed the need for the CAF_WORKER_DATASTORE_NO_DELETE_DIRS cfg completely, the rabbit consumer passes the noDelete directory on calling delete.

dermot-hardy (approved):

worker-store-fs/src/main/java/com/github/workerframework/datastores/fs/FileSystemDataStore.java
```diff
@@ -101,23 +102,45 @@ public void shutdown()
...
+     * Delete a Directory tree from leaf node up if cleanPartialReference is true until a non-empty directory is encountered. If
+     * cleanPartialReference is false only the referenced file is deleted.
+     *
+     * @param reference the file to be deleted, along with non-empty parent directories in the files directory tree.
+     * @param cleanPartialReference whether the partial reference area should be cleared out
+     * @param noDeleteDirectory a directory in the reference where the cleanup should stop
+     * @throws DataStoreException if the directory cannot be accessed or deleted
+     */
+    @Override
+    public void delete(final String reference, final boolean cleanPartialReference, final String noDeleteDirectory) throws DataStoreException
```
dermot-hardy:
```suggestion
    public void delete(final String reference, final boolean cleanPartialReference, final String noDeleteDirectory)
        throws DataStoreException
```

worker-store-fs/src/main/java/com/github/workerframework/datastores/fs/FileSystemDataStore.java
```diff
@@ -101,23 +102,45 @@ public void shutdown()
...
+    {
+        Objects.requireNonNull(reference);
+        final List<Path> noDeletePaths = List.of(dataStorePath, dataStorePath.resolve(noDeleteDirectory));
+        delete(reference, cleanPartialReference, noDeletePaths);
+    }
+
+    private void delete(final String reference, final boolean cleanPartialReference, final List<Path> noDeletePaths) throws DataStoreException
```
dermot-hardy:
```suggestion
    private void delete(final String reference, final boolean cleanPartialReference, final List<Path> noDeletePaths)
        throws DataStoreException
```

